PR #37, opened 2014-10-29
Date based partitioning

tzaavi:
Added functionality to the DateMessageParser to handle partitioning data into a date-based directory structure. For example, the user can specify
a partition pattern that writes files to the following folder structure:

s3n://.../y=2014/m=08/d=09/...

This change maintains backwards compatibility with the original
partitioning scheme with prefix.


silasdavis:
Can we close this pull request to tidy up?

tzaavi:
yes

PR #60, opened 2014-12-03
Changed the FileReaderWriter class to a factory based implementation

silasdavis:
The more I thought about the FileReaderWriter implementation the more I thought it was an example of stringly typing (only with a enum). I really think it is a bit of a code smell so I refactored it into a Factory-based implementation.

This maintains the ability to specifiy a single factory class that generates readers and writers (see the SequenceFileReaderWriterFactory class). The standard way to implement would be to provide the reader and writer implementation as nested classes as I have done with the existing SequenceFile and DelimteredTextFile ones.

I've made all the tests pass, but I'm sure there's some cleaning up to do. I just wanted to show it to you early to see what you think.

Unavoidable its a big PR, but the main entry point is: src/main/java/com/pinterest/secor/io/FileReaderWriterFactory.java

All changes cascade from that. The nice thing about this is you have the intialisation method directly on the factory, so implementers know exactly what to implement and the type system enforces this. The only reflection is to create the factory (which requires a parameterless constructor).


silasdavis:
OK I've removed the default Kafka values and made the factory creation reflection method private. Tests passed and I've rebased against master.

pgarbacki:
@silasdavis this is good for merging but would you mind squashing your commits?

silasdavis:
I've just squashed the commits and rebased off master

pgarbacki:
Thanks @silasdavis for this valuable contribution and sorry again about the long turn around time. Feel free to add yourself to the contributors in README.md.

PR #64, opened 2015-01-05
Adding support for finalizing topics that have json messages

ramkivenkatachalam:
Adding support for optional prefix for hive tables
Fixing PartitionFinalizer to work for json topics.


ramkivenkatachalam:
Thanks, pawel.

PR #65, opened 2015-01-08
Generalizing number conversion

barata:
This approach works for integer, float/double and scientific notation (e.g., 12E+9).
The throws declaration was removed from method extractTimestampMillis().


pgarbacki:
thanks for the contribution, @barata

PR #71, opened 2015-01-16
using length of BytesWritable to decide how much of the byte array to di...

estezz:
This is a know hadoop issue.  It is discussed here This is the bug in Hadoop https://issues.apache.org/jira/browse/HADOOP-6298.

Now using this to get the bytes out.
System.arraycopy(value.getBytes(), 0, nonPaddedBytes, 0, value.getLength());


pgarbacki:
Thanks @estezz . Feel free to add yourself to the list of contributors unless you are already there.

PR #74, opened 2015-01-28
output dir structure partitioning based off on a config param

nkunal:
The output dir structure defaults to dt=yyyy-MM-dd, this pull request will make this configurable on how the date formatting needs to be done.

### DIFF
--- a/src/main/java/com/pinterest/secor/parser/TimestampedMessageParser.java
+++ b/src/main/java/com/pinterest/secor/parser/TimestampedMessageParser.java
@@ -26,11 +26,18 @@ import java.util.TimeZone;
 public abstract class TimestampedMessageParser extends MessageParser {
 
     private SimpleDateFormat mFormatter;
+    Object outputPattern = null;
 
     public TimestampedMessageParser(SecorConfig config) {
         super(config);
-        mFormatter = new SimpleDateFormat("yyyy-MM-dd");
-        mFormatter.setTimeZone(TimeZone.getTimeZone("UTC"));
+	outputPattern = mConfig.getOutputPattern();
+	if (outputPattern == null) {
+	    mFormatter = new SimpleDateFormat("yyyy-MM-dd");
+	    mFormatter.setTimeZone(TimeZone.getTimeZone("UTC"));
+	} else {
+	    mFormatter = new SimpleDateFormat(outputPattern.toString());
+	    mFormatter.setTimeZone(TimeZone.getTimeZone("EST"));
+	}
     }
 
     public abstract long extractTimestampMillis(final Message message) throws Exception;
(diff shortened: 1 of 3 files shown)


src/main/java/com/pinterest/secor/parser/TimestampedMessageParser.java
```diff
@@ -26,11 +26,18 @@
...
         super(config);
-        mFormatter = new SimpleDateFormat("yyyy-MM-dd");
-        mFormatter.setTimeZone(TimeZone.getTimeZone("UTC"));
+	outputPattern = mConfig.getOutputPattern();
+	if (outputPattern == null) {
+	    mFormatter = new SimpleDateFormat("yyyy-MM-dd");
+	    mFormatter.setTimeZone(TimeZone.getTimeZone("UTC"));
+	} else {
+	    mFormatter = new SimpleDateFormat(outputPattern.toString());
+	    mFormatter.setTimeZone(TimeZone.getTimeZone("EST"));
```
zackdever: timezone shouldn't be set to EST because a different date format is specified

src/main/java/com/pinterest/secor/parser/TimestampedMessageParser.java
```diff
@@ -26,11 +26,18 @@
 public abstract class TimestampedMessageParser extends MessageParser {
 
     private SimpleDateFormat mFormatter;
+    Object outputPattern = null;
```
zackdever: Think this ought to be a `String`, but according to https://github.com/pinterest/secor/pull/10#discussion_r15147319 I think the style in this project is to just access the config inline as needed.

zackdever:
small detail - while a few tabs have snuck in, it would be nice to convert these to spaces

src/main/java/com/pinterest/secor/parser/TimestampedMessageParser.java
nkunal: yup, fixed it in a new commit

nkunal:
incorporate feedback from zackdever, ready for further review/merge

pgarbacki:
Thanks @nkunal for the contribution. In general, I think it is dangerous to expose the ability of arbitrary formatting of the date partition. It is quite common for secor output to be imported into Hive (in fact secor comes with utilities that automatically finalize day partitions). Hive does some wired things converting special characters (such as directory separators) to partition names and I would like to avoid the possibility of users shooting themselves in the foot. Also, to decrease the maintenance overhead, I would prefer to minimize the number of configuration params that Secor exposes.

Note that you can still support your use case by implementing a custom MessageParser. In fact, the parsers are intended to be extended rather that customized.

pgarbacki:
BTW. now that we have secor-contrib repo, if you decide to implement a custom parser that you find generally useful, you may consider putting it there.

nkunal:
@pgarbacki ok cool, I will go ahead with the custom parser. 
Another related question, currently the output directory structure is s3n://bucket/topic/<partitioning_logic>
If I have to change this to s3n://bucket/topic_<some_value>/<partitioning_logic>, inorder to achieve this, the customer parser would not suffice. The basename is handled in LogFilePath.java, what would be a preferable way to do this ? so that it does not cause harm for other use cases.

pgarbacki:
@nkunal changing the topic name in the path component is not supported. Customizations at this level are a bit tricky. In particular, configuration errors could result in overwriting logs.

HenryCaiHaiying:

PR #75, opened 2015-02-08
Add support for additional Kafka fetch options

mthssdrbrg:
Quite useful options for keeping the number of requests to a minimum for topics that doesn't get `fetch.message.max.bytes` worth of data within the default `fetch.wait.max.ms`, which is 100ms.


pgarbacki:
Sorry about the delay.

PR #76, opened 2015-02-08
(Fully) local integration tests

mthssdrbrg:
This introduces a way to run the integration tests in a purely local fashion, using `fakes3` (a Ruby gem) rather than an actual S3 bucket.

In order to accomplish this a few configuration options needs to be set (AWS keys and S3 bucket), and I’ve included “defaults” that will work on Travis (in separate config. files). One also needs to add an entry to `/etc/hosts` (or equivalent), that points `test-bucket.localhost` to `127.0.0.1`.

There's a rather naive function, `check_for_native_libs`, that checks whether `HADOOP_NATIVE_LIB_PATH` actually contains any `*.so` files, and if it doesn’t it’ll ignore the “compressed tests” for sequence files, as it's not possible to run them without the native libs, and it's pretty darn difficult to build them on Mac OS X, not even sure if it's supported at all.

Furthermore there’s a Makefile that simplifies running the integration tests, a simple `make integration` is all that’s needed. It’ll build, package and extract Secor to a directory in `/tmp` and then run the tests.

Perhaps there should be a section in the README dedicated to how the test setup works, and how to run the tests.

Let me know what you think of this. The main benefit I see from this is having each pull request automatically run the integration tests on Travis (given that it's enable that is), and for me it's nice to be able to run the integration tests locally without having to communicate with S3 or any other third-party service (though it's only S3 in this case).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,20 @@
 language: java
+addons:
+  hosts:
+    - test-bucket.localhost
+env:
+  - PATH=$PATH:$HOME/.s3cmd SECOR_LOCAL_S3=true S3CMD=1.0.1
 jdk:
-  - oraclejdk8
-  - oraclejdk7
-  - openjdk6
   - openjdk7
+  - oraclejdk7
+  - oraclejdk8
+before_install:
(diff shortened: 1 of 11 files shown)


pgarbacki:
@mthssdrbrg this is truly awesome. Thanks so much for making Secor easier to test!

PR #79, opened 2015-03-04
parameterized broker

mulloymorrow:
useful to specify broker in this test


src/main/java/com/pinterest/secor/tools/TestLogMessageProducer.java
```diff
@@ -40,16 +40,18 @@
     private final String mTopic;
     private final int mNumMessages;
     private final String mType;
+    private String brokerList = "localhost:9092";
```
pgarbacki: For stylistic consistency, you should inline it.

pgarbacki:
@mulloymorrow could you address the comments and rebase?

PR #87, opened 2015-04-23
Use s3a:// filesystem instead of s3n://.

kecaps:
Removed jets3t dependencies.
In run_tests.sh, quote variable in -n test.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,11 +49,6 @@
 
     <dependencies>
         <dependency>
-            <groupId>net.java.dev.jets3t</groupId>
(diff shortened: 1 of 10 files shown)


pgarbacki:
I'm not very familiar with the s3a filesystem. From what I read online ( https://wiki.apache.org/hadoop/AmazonS3 ), s3a is backwards compatible with s3n, but s3a files may not be readable through s3n urls. We have plenty of downstream deps on s3n (hive tables, hadoop jobs, etc.) that will break if we change the url. For now I would prefer to stick with the wider supported fs.

Issue #41, opened 2014-11-05
Support Avro Message Parser


upio:
Secor only supports Thrift and JSON messages out of the box right now. Secor should provide a configurable Avro Message Parser which uses Avros GenericRecord parser and a configurable Timestamp field.

PR #90, opened 2015-05-05
Support Avro Message Parser

ankurcha:
This pull request adds avro message parser support which is similar to the Thrift date parser. This change adds the setting: `message.input.avro.schema` which corresponds to the path of the avro schema file.

Associated issue #41


ppearcy:
FYI, I have some work underway to support confluent schema repo for avro:
https://github.com/pinterest/secor/pull/114

I'm not sure how well the changes in this PR would work as the generated avro files wouldn't include the schema.

ankurcha:
That would be ideal but i don't see everyone using the schema registry and this change is to maintain the isolated operational dependencies of secor.

jlmichael:
What is the status of this PR?  Is it still being worked on?  Sounds like exactly what I want.  :)

pgarbacki:
@HenryCaiHaiying will take a look at this PR. In the meantime, could someone fix the tests?

README.md
```diff
@@ -53,6 +53,8 @@ One of the convenience features of Secor is the ability to group messages and sa
...
 
+- **avro date parser**: parser that extracts timestamps from avro messages and groups the output based on the date (at a day granularity). To keep things simple, this parser assumes that the timestamp is carried in the 'timestamp' field of the avro message schema. The timestamp may be expressed either in seconds or milliseconds, or nanoseconds since the epoch. The output goes to date-partitioned paths (e.g., ```s3n://bucket/topic/dt=2014-05-01```, ```s3n://bucket/topic/dt=2014-05-02```). Date pertitioning is particularly convenient if the output is to be consumed by ETL tools such as [Hive]. To use this parser, start Secor with properties file [secor.prod.partition.properties](src/main/config/secor.prod.partition.properties). You may override the field used to extract the timestamp by setting the "message.timestamp.name" property. The schema file for the avro messages must be provided using the setting "message.input.avro.schema".
```
HenryCaiHaiying: Instead of copying the description from 'thrift date parser', can you just refer to it?

pom.xml
```diff
@@ -165,17 +165,17 @@
         <dependency>
             <groupId>org.apache.thrift</groupId>
             <artifactId>libthrift</artifactId>
-            <version>0.5.0</version>
+            <version>0.9.2</version>
```
HenryCaiHaiying: This is a big version jump.  Are these version upgrade needed to support avro?

src/main/java/com/pinterest/secor/parser/AvroMessageParser.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public class AvroMessageParser extends TimestampedMessageParser {
+    final DatumReader<GenericRecord> mDatumReader;
+    final String mFieldName;
+
+    public AvroMessageParser(SecorConfig config) throws IOException {
+        super(config);
+        Schema schema = new Schema.Parser().parse(new File(mConfig.getMessageInputAvroSchema()));
+        mFieldName = config.getMessageTimestampName();
+        mDatumReader = new GenericDatumReader<GenericRecord>(schema);
```
HenryCaiHaiying: If there are multiple topics being pulled in secor, you would need to different schemas for different topics, looks like this  config might need to be a map.

src/main/java/com/pinterest/secor/parser/AvroMessageParser.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    public long extractTimestampMillis(Message message) throws Exception {
+        byte[] payload = message.getPayload();
+        if (payload == null || payload.length == 0) {
+            return 0L;
```
HenryCaiHaiying: Do you want to add a logging in this case?

src/main/java/com/pinterest/secor/parser/AvroMessageParser.java
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        BinaryDecoder decoder = DecoderFactory.get().binaryDecoder(payload, null);
+        GenericRecord record = mDatumReader.read(null, decoder);
```
HenryCaiHaiying: What if reader cannot pass the data record (e.g. schema mismatch)?

src/main/java/com/pinterest/secor/parser/AvroMessageParser.java
```diff
@@ -0,0 +1,47 @@
...
+        Object val = record.get(mFieldName);
+        Long longVal = Longs.tryParse(String.valueOf(val));
```
HenryCaiHaiying: You need to deal with error case better (field is not there, type is not a Long etc)

HenryCaiHaiying:
In general, reading the avro schema from a file is not as robust as reading it from schema registry.

HenryCaiHaiying:

PR #94, opened 2015-05-15
adding stats counter for rebalance

ramkivenkatachalam:
adding stats counter for kafka consumer rebalance


pgarbacki:
r+ if tests pass

PR #95, opened 2015-05-20
avoid message toString when debug not enabled

yyejun:
message.toString is called twice even when debug logging is not enabled

Because message can be huge, I used this method to avoid unnecessary calls. 
http://www.slf4j.org/faq.html#logging_performance


pgarbacki:
@yyejun thanks for the contribution!

PR #96, opened 2015-05-23
Fix memory leak when DelimitedTextFileReaderWriterFactory combined with native lib

yyejun:
This fix is borrowed from org.apache.hadoop.io.SequenceFile.

This will also cause extra log line "INFO  Got brand-new decompressor [.gz]" when native lib not enabled. This behavior should be safe as builtin gzip decompressor is not cacheable.


pgarbacki:
Thanks for the contribution, @yyejun

PR #98, opened 2015-06-06
Support s3a and s3n filesystems based on configuration property.

kecaps:
This also contains some enhancements to LogFilePrinter

### DIFF
--- a/src/main/config/secor.common.properties
+++ b/src/main/config/secor.common.properties
@@ -28,6 +28,18 @@ aws.secret.key=
 # END MUST SET #
 ################
 
+# AWS region or endpoint. region should be a known region name (eg.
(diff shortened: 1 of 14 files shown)


src/main/config/secor.common.properties
```diff
@@ -28,6 +29,10 @@ aws.secret.key=
 # END MUST SET #
 ################
 
+# Hadoop filesystem to use. Choices are s3n or s3a.
+# See https://wiki.apache.org/hadoop/AmazonS3 for details.
+secor.s3.filesystem=s3a
```
pgarbacki: Could you change the default to s3n to preserve backwards compatibility?

pgarbacki:
@kecaps could you rebase and squash the commits?

PR #99, opened 2015-06-09
performance fix for small messages

yyejun:
This specific property is reloaded on every message read which cause server performance limited around 20k/s. Cache this as a static value.

### DIFF
--- a/src/main/java/com/pinterest/secor/reader/MessageReader.java
+++ b/src/main/java/com/pinterest/secor/reader/MessageReader.java
@@ -54,6 +54,9 @@ public class MessageReader {
     private ConsumerConnector mConsumerConnector;
     private ConsumerIterator mIterator;
     private HashMap<TopicPartition, Long> mLastAccessTime;
+    private final int mTopicPartitionForgetSeconds;
+    private final int mCheckMessagesPerSecond;
+    private int mNMessages;
 
     public MessageReader(SecorConfig config, OffsetTracker offsetTracker) throws
             UnknownHostException {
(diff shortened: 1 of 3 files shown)


pgarbacki:
r+

yyejun:
* remove unnecessary buffer
* only rate limit and export stats every seconds or so

src/main/java/com/pinterest/secor/reader/MessageReader.java
```diff
@@ -54,6 +54,9 @@
     private ConsumerConnector mConsumerConnector;
     private ConsumerIterator mIterator;
     private HashMap<TopicPartition, Long> mLastAccessTime;
+    private final int mTopicPartitionForgetSeconds;
+    private final int checkMessagesPerSecond;
```
pgarbacki: This should be called mCheckMessagesPerSecond.

src/main/java/com/pinterest/secor/reader/MessageReader.java
```diff
@@ -54,6 +54,9 @@
...
+    private int nMessages;
```
pgarbacki: mNMessages

src/main/java/com/pinterest/secor/reader/MessageReader.java
```diff
@@ -141,7 +146,9 @@ public boolean hasNext() {
 
     public Message read() {
         assert hasNext();
-        RateLimitUtil.acquire();
+        if (nMessages++ % checkMessagesPerSecond == 0) {
```
pgarbacki: nMessages = (nMessages + 1) % checkMessagesPerSecond
to prevent overflow

src/main/java/com/pinterest/secor/util/RateLimitUtil.java
```diff
@@ -35,6 +35,10 @@ public static void configure(SecorConfig config) {
         mRateLimiter = RateLimiter.create(config.getMessagesPerSecond());
     }
 
+    public static void acquire(int n) {
+        mRateLimiter.acquire(n);
+    }
+
     public static void acquire() {
```
pgarbacki: Is this method still in use?

pgarbacki:
Please squash your commits.

PR #102, opened 2015-06-30
Fix DelimitedTextFile offset.

### LINKED ISSUE #101: DelimitedTextFile trim file incorrectly around rebalance
DelimitedTextFile assumed LogFilePath's offset is the starting offset, and entire file is continuous stream. 

First problem will be a rebalance happened after datetime partition switch, but this doesn't happen very often at same time with a rebalance event.

Second situation is an old thread picked a topic again after a period time, it will still use previous committed offset as filename which can be days ago.

yyejun:
Use a sequence file to manage offset.
fix #101

### DIFF
--- a/src/main/java/com/pinterest/secor/common/FileRegistry.java
+++ b/src/main/java/com/pinterest/secor/common/FileRegistry.java
@@ -17,6 +17,7 @@
 package com.pinterest.secor.common;
 
 import com.pinterest.secor.io.FileWriter;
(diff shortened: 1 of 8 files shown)


pgarbacki:
@yyejun would you mind rebasing?

yyejun:
I rebased, but not sure what caused tests exceed 50 minutes.

HenryCaiHaiying:

PR #103, opened 2015-07-01
Support uploads using the AWS SDK TransferManager.

liamstewart:
This provides more flexibility for credentials and region choice. The
core upload functionality has been moved out of `Uploader` and into
`UploadManager` implementations. The manager to use can be set through
the configuration file.

Similar to https://github.com/pinterest/secor/pull/98 but without the Hadoop layer.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -92,6 +92,11 @@
             <version>2.6.1</version>
         </dependency>
         <dependency>
+            <groupId>com.amazonaws</groupId>
+            <artifactId>aws-java-sdk-s3</artifactId>
+            <version>1.10.2</version>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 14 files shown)


nguyenvanthan:
Hi @liamstewart ,
Why don't you use only __aws-java-sdk-s3__ artifact instead of full sdk ?

liamstewart:
Yes, the depedency really should be on the s3 part of the sdk, rather than the full sdk. I'll update.

pgarbacki:
@liamstewart could you rebase your change and squash the commits?

liamstewart:
Rebased and squashed.

pgarbacki:
Thanks for the contribution @liamstewart

PR #106, opened 2015-07-02
Fix a race condition when two upload happens at the same time

### LINKED ISSUE #105: Data loss during rebalance
We observed a data loss during rebalance. This happens very rarely, but look at code path, this definitely can happen.

The current consumer deleteWriters, then another thread committed offset. 
https://github.com/pinterest/secor/blob/master/src/main/java/com/pinterest/secor/uploader/Uploader.java#L96

https://github.com/pinterest/secor/blob/master/src/main/java/com/pinterest/secor/uploader/Uploader.java#L102

This if statement return false, nothing happens, All data in this writer will be lost.

Since this is a very rare case, I think maybe the best solution will be throw a runtime exception if the if statement return false?

yyejun:
Fix #105

### DIFF
--- a/src/main/java/com/pinterest/secor/uploader/Uploader.java
+++ b/src/main/java/com/pinterest/secor/uploader/Uploader.java
@@ -102,8 +102,6 @@ public class Uploader {
             topicPartition.getTopic(),
             topicPartition.getPartition());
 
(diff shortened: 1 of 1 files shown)


pgarbacki:
These seem to be some conflicts preventing this PR from being merged. Could you resolve those?

yyejun:
rebased

PR #108, opened 2015-07-09
avoid string concatenation in slf4j logs

nguyenvanthan:
Use slf4j api.
Remove string concatenation (string additional objects on runtime).


src/main/java/com/pinterest/secor/common/FileRegistry.java
```diff
@@ -102,7 +102,7 @@ public FileWriter getOrCreateWriter(LogFilePath path, CompressionCodec codec)
             writer = ReflectionUtil.createFileWriter(mConfig.getFileReaderWriterFactory(), path, codec);
             mWriters.put(path, writer);
             mCreationTimes.put(path, System.currentTimeMillis() / 1000L);
-            LOG.debug("created writer for path " + path.getLogFilePath());
+            LOG.debug("created writer for path {}", path.getLogFilePath());
```
yyejun: getLogFilePath should not be called here if you want any performance benefit from this change. toString is method of path is same as getLogFilePath.

PR #109, opened 2015-07-14
    Adding support for hourly s3 data ingestion from secor:

HenryCaiHaiying:
Summary:
    - Add the capability of partition the data file in S3 using hourly folder:
       s3://pinlogs/secor_raw_logs/topic1/dt=2015-07-07/hr=05/
      This way, the data file will be on S3 much sooner (hourly vs daily), people can check the result on S3 much sooner and this also opens the door to have an faster hourly data pipeline on Hadoop side as well.
      The hr folder values are from 00-23

    - To trigger the hourly partition and finalization, add the following parameter in your secor config:
        * message.timestamp.using.hour=true
      And change the upload threshold to less than one hour:
        * secor.max.file.age.seconds=3000

    - Change the hive partition registration code to register partition using both dt and hr column (it does require the HIVE table to be created or altered to have both dt and hr as the partition columns)

    - The enhancements are done through the following:
      * Change the TimestampedMessageParser to allow it to extract both the dt and hr from the timestamp field
      * Change the Finalizer to go through both the hr and dt folder to generate the SUCCESS file and do hive registration
        - the dt finalizer lag before finalization was old way: 1 day + 1 hour lag
        - the hr finalizer lag was defined as: 1 hour lag

    - Added more unit test on message parser on hourly behavior
    - Added more E2E tests to cover the partitioner and hourly ingestion

    Test Plan: Added both unit tests and e2e tests, and tested manually for a new topic on S3

### DIFF
--- a/src/main/config/secor.dev.hr.partition.properties
+++ b/src/main/config/secor.dev.hr.partition.properties
@@ -1,0 +1,11 @@
+include=secor.dev.properties
+
+secor.kafka.group=secor_hr_partition
+secor.message.parser.class=com.pinterest.secor.parser.ThriftMessageParser
+
+secor.s3.path=secor_dev/hr_partition
+secor.local.path=/tmp/secor_dev/message_logs/hr_partition
+
+message.timestamp.using.hour=true
+
+ostrich.port=9998
--- a/src/main/config/secor.dev.hr.partition.properties
+++ b/src/main/config/secor.dev.hr.partition.properties
@@ -6,6 +6,6 @@ secor.message.parser.class=com.pinterest.secor.parser.ThriftMessageParser
 secor.s3.path=secor_dev/hr_partition
 secor.local.path=/tmp/secor_dev/message_logs/hr_partition
 
-message.timestamp.using.hour=true
+partitioner.granularity.hour=true
 
 ostrich.port=9998
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -215,6 +215,14 @@ public class SecorConfig {
       return getBoolean("message.timestamp.using.hour", false);
     }
 
+    public int getFinalizerLagSecond() {
+        return getInt("finalizer.lag.second", 3600);
+    }
+
+    public int getFinalizerLookbackPeriods() {
+        return getInt("finalizer.lookback.periods", 10);
+    }
+
     public String getHivePrefix() { 
         return getString("secor.hive.prefix"); 
     }
@@ -256,12 +264,12 @@ public class SecorConfig {
         return mProperties.getInt(name);
     }
 
-    private boolean getBoolean(String name, boolean required) {
-      if (required) {
-        checkProperty(name);
-        return mProperties.getBoolean(name);
-      }
-      return mProperties.getBoolean(name, false);
+    private int getInt(String name, int defaultValue) {
+        return mProperties.getInt(name, defaultValue);
+    }
+
+    private boolean getBoolean(String name, boolean defaultValue) {
+        return mProperties.getBoolean(name, defaultValue);
     }
 
     private long getLong(String name) {
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -211,16 +211,8 @@ public class SecorConfig {
         return getString("message.timestamp.input.pattern");
     }
 
-    public boolean getMessageTimestampUsingHour() {
-      return getBoolean("message.timestamp.using.hour", false);
-    }
-
-    public int getFinalizerLagSecond() {
-        return getInt("finalizer.lag.second", 3600);
-    }
-
     public int getFinalizerLookbackPeriods() {
-        return getInt("finalizer.lookback.periods", 10);
+        return getInt("secor.finalizer.lookback.periods", 10);
     }
 
     public String getHivePrefix() { 
@@ -247,6 +239,10 @@ public class SecorConfig {
         return getString("secor.zookeeper.path");
     }
 
+    public boolean getBoolean(String name, boolean defaultValue) {
+        return mProperties.getBoolean(name, defaultValue);
+    }
+
     private void checkProperty(String name) {
         if (!mProperties.containsKey(name)) {
             throw new RuntimeException("Failed to find required configuration option '" +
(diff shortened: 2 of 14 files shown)


src/main/config/secor.dev.hr.partition.properties
```diff
@@ -0,0 +1,11 @@
+include=secor.dev.properties
+
+secor.kafka.group=secor_hr_partition
+secor.message.parser.class=com.pinterest.secor.parser.ThriftMessageParser
+
+secor.s3.path=secor_dev/hr_partition
+secor.local.path=/tmp/secor_dev/message_logs/hr_partition
+
+message.timestamp.using.hour=true
```
pgarbacki: We should limit the number of params to bare minimum. Let's introduce a separate parser class for hourly partitions.

src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -211,6 +211,14 @@ public String getMessageTimestampInputPattern() {
         return getString("message.timestamp.input.pattern");
     }
 
+    public boolean getMessageTimestampUsingHour() {
+      return getBoolean("message.timestamp.using.hour", false);
+    }
+
+    public int getFinalizerLookbackPeriods() {
+        return getInt("finalizer.lookback.periods", 10);
```
pgarbacki: secor.finalizer.lookback.period

src/main/java/com/pinterest/secor/parser/PartitionFinalizer.java
```diff
@@ -61,154 +60,123 @@ public PartitionFinalizer(SecorConfig config) throws Exception {
...
-            }
-        }
-        if (minTimestamp == Long.MAX_VALUE) {
-            return -1;
-        }
-        return minTimestamp;
-    }
+        // Walk through each dimension of the partitions array
+        // First finalize the hourly partitions, and then move on to daily partitions
+        for (int dim = partitions.length; dim > 0; dim--) {
```
pgarbacki: I think it would be cleaner and more generic to let the Partitioner decide on the ordering. E.g., the partitioner could propose the following order:

dt=2015-07-01
dt=2015-07-01/hr=01
dt=2015-07-01/hr=02
...
dt=2015-07-01/hr=23
dt=2015-07-02
dt=2015-07-02/hr=01
...

pgarbacki:
I did a quick swipe through the code. The implementation looks clean and nicely structured. Also, great documentation and thorough tests.

Please squash your commits and feel free to add yourself to the list of contributors.

src/main/config/secor.dev.hr.partition.properties
HenryCaiHaiying: There are already several several sub-classes of TimestampedMessageParser (json, thrift parser e.g.) and they can all benefit from hourly parsing/partitioning.   if I introduce another hourly parser, I would have to duplicate code between those parser classes.

If the timestamped parser understands the daily concept, it would be natural for it to understand hourly concept too.

src/main/java/com/pinterest/secor/common/SecorConfig.java
HenryCaiHaiying: OK

src/main/java/com/pinterest/secor/parser/PartitionFinalizer.java
HenryCaiHaiying: If we do this, this is probably better done to have Partitioner to provide a Iterator class or simply returns a list of those partitions (bounded by look.back.periods), otherwise it would be hard for partitioner to keep the iterating state.

There are some differences between those partitions, e.g. dt=2015-07-01 don't need to do hive registration while the hourly partition needs to, I guess the finalizer needs to figure this out maybe by comparing the length of the those partition elements.

src/main/java/com/pinterest/secor/parser/PartitionFinalizer.java
```diff
@@ -41,12 +39,13 @@
...
 
-    private SecorConfig mConfig;
-    private ZookeeperConnector mZookeeperConnector;
-    private TimestampedMessageParser mMessageParser;
-    private KafkaClient mKafkaClient;
-    private QuboleClient mQuboleClient;
-    private String mFileExtension;
+    private final SecorConfig mConfig;
+    private final ZookeeperConnector mZookeeperConnector;
+    private final TimestampedMessageParser mMessageParser;
```
pgarbacki: Its fine for now but in the future we may want to make it easier to plug different message parsers and partitioners into the finalizer.

src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -211,6 +211,10 @@ public String getMessageTimestampInputPattern() {
         return getString("message.timestamp.input.pattern");
     }
 
+    public int getFinalizerLookbackPeriods() {
+        return getInt("secor.finalizer.lookback.periods", 10);
```
pgarbacki: You should put this param in one of the .properties files and add some documentation. You can comment it out by default.

src/main/config/secor.dev.hr.partition.properties
pgarbacki: In the spirit of separating parameter namespaces, could you rename it to
partitioner.granularity.hour
or something like that?

src/main/java/com/pinterest/secor/parser/TimestampedMessageParser.java
```diff
@@ -49,12 +72,101 @@ protected static long toMillis(final long timestamp) {
         return timestampMillis;
     }
 
+    protected String[] generatePartitions(long timestampMillis, boolean usingHourly)
```
pgarbacki: Why do you need usingHourly param in addition to the mUsingHourly attribute?

src/main/java/com/pinterest/secor/common/SecorConfig.java
HenryCaiHaiying: Will do.

src/main/config/secor.dev.hr.partition.properties
HenryCaiHaiying: OK.

src/main/java/com/pinterest/secor/parser/TimestampedMessageParser.java
HenryCaiHaiying: In the hourly partitioner, we are generating both hourly and daily partition string.  So the caller of this method needs to decide whether for a particular partition string, whether a daily format or a hourly format should be used (even though mUsingHourly is always true for that partitioner).

pgarbacki:
r+

Please squash the commits before merging. Also, add yourself to the list of contributors.

PR #111, opened 2015-07-15
Simple timestamped message parser for protocol buffers.

liamstewart:
This is similar to the thrift message parser in that it expects the first field to be a uint64 timestamp. Works well enough for us and avoids either having to create a parser that either links in generated protobuf code or uses dynamic messages and descriptors.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,11 @@
 
     <dependencies>
         <dependency>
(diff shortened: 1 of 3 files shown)


pgarbacki:
@liamstewart thanks for the contribution

PR #114, opened 2015-07-24
Support params for schema registry and add optional parsedRecord to KeyValue to allow more context to be passed to writer/reader

ppearcy:
This PR isn't complete, but before I went further wanted to make sure maintainers were on board with the change and open things up to discussion. 

It pushes a schema registry into both the writer and the parser to enable messages to be parsed coming out of Kafka and written out with the proper avro object container format. 

Things that are left to do before I'd expect this to be mergeable:
- Tests, duh
- Support compression
- Port over avro message parser to Java and include it: 
  https://gist.github.com/ppearcy/d6aee9be77cd9aa362a3

There's also the potential to plugin other schema registries, similar as other components. I'm not quite sure if that makes sense yet, unless you know of other registries out there you know you'd want to support. 

These changes also open up the door for Secor to read avro out of Kafka and generate either parquet (via AvroSchemaConverter) or to orc file (via  AvroObjectInspectorGenerator).

### DIFF
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -112,11 +112,11 @@ public class SecorConfig {
     }
 
     public String getSchemaUrl() {
-        return getString("schema.url");
+        return getString("schema.registry.url");
     }
 
     public int getSchemaCacheCount() {
-        return getInt("schema.cache.count", 1000);
+        return getInt("schema.registry.cache.count", 1000);
     }
 
     public int getGeneration() {
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -111,6 +111,14 @@ public class SecorConfig {
         return getString("kafka.fetch.wait.max.ms");
     }
 
+    public String getSchemaUrl() {
+        return getString("schema.url");
+    }
+
+    public int getSchemaCacheCount() {
+        return getInt("schema.cache.count", 1000);
+    }
+
     public int getGeneration() {
         return getInt("secor.generation");
     }
--- a/src/main/config/secor.common.properties
+++ b/src/main/config/secor.common.properties
@@ -168,7 +168,7 @@ secor.upload.manager.class=com.pinterest.secor.uploader.HadoopS3UploadManager
 
 # Schema registry URL that can be used by parsers that support or require a schema registry
 # This can be used for formats that require the schema to deserialize, eg Avro
-schema.url=
+schema.registry.url=
 
 # Number of cached schemas the registry should keep local at a time. Defaults to 1000.
-schema.cache.count=
+schema.registry.cache.count=
--- a/src/main/config/secor.common.properties
+++ b/src/main/config/secor.common.properties
@@ -165,3 +165,10 @@ secor.max.message.size.bytes=100000
 # Class that will manage uploads. Default is to use the hadoop
 # interface to S3.
 secor.upload.manager.class=com.pinterest.secor.uploader.HadoopS3UploadManager
+
+# Schema registry URL that can be used by parsers that support or require a schema registry
+# This can be used for formats that require the schema to deserialize, eg Avro
+schema.url=
+
+# Number of cached schemas the registry should keep local at a time. Defaults to 1000.
+schema.cache.count=
--- a/src/main/java/com/pinterest/secor/io/KeyValue.java
+++ b/src/main/java/com/pinterest/secor/io/KeyValue.java
@@ -27,13 +27,21 @@ public class KeyValue {
 	
 	private final long mKey;
 	private final byte[] mValue;
-	
+	private final Object mParsedRecord;
+
 	//constructor
 	public KeyValue(long key, byte[] value) {
 		this.mKey = key;
 		this.mValue = value;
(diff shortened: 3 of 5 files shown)


ppearcy:
This would likely address: https://github.com/pinterest/secor/issues/41

HenryCaiHaiying:
Some high level comment:

- SchemaRegistry is only needed for Avro path, not eveyone needs the avro schema, SchemaRegistry should not be a top level concept into Consumer, MessageParser etc, I think you should be able to capture the SchemaRegistry within AvroFileReaderWriterFactory, similarly you can create a sublcass of MessageParser (AvroMessageParser, similar to ThriftMessageParser) to parse the avro message
- How do you handle the avro schema migration, e.g. the schema for the later records changed.  You are using the schema for the first record as the write schema for the avro sequence file, do you want the latest schema instead?  I think Avro schema registry also have the concept of schema per type, maybe you can use that schema as the write schema for avro file.
- We need unit test and integration test coverage for avro path

ppearcy:
Thanks for the feedback. Regarding having avro registry a top level component, in general, you only want one instance as it's stateful/has a cache. However, I figured out how to only need that instance down in the MessageParser and using the avro container format (including the schema) to pass along to the writer. 

Regarding schema migration, you are correct, if I wanted to, I could pull the latest schema and use that to write with. It isn't critical, though, as we can specify the schema when we read the data back out of S3 and have migration kick in there. It would be nice, but I'd like to see that baked into the kafka schema registry client and opened a ticket for that: https://github.com/confluentinc/schema-registry/issues/216

With the recent updates, I have removed almost all my core changes to Secor. The only thing that is left is the extra config properties. It would be nice not to have core app know about all config properties every parser/writer/etc needs. 

A few questions that would help me progress:
* Do you think it makes sense to allow generic properties that can flow through (this should be easy to expose) 
* Is the Avro schema registry support something you want in core secor? 
* Are there objections to the scala code I added (if so, not a big deal, it should be minor to port to Java).

pom.xml
```diff
@@ -385,6 +399,11 @@
                     <autoReleaseAfterClose>true</autoReleaseAfterClose>
                 </configuration>
             </plugin>
+            <plugin>
```
HenryCaiHaiying: Please don't introduce the scala dependency, it's hard to maintain scala version upgrade.  Even the kafka consumers are re-written in Java now.

src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -111,6 +111,14 @@ public String getFetchWaitMaxMs() {
         return getString("kafka.fetch.wait.max.ms");
     }
 
+    public String getKafkaSchemaUrl() {
```
HenryCaiHaiying: 1. I though each time you add a new parameter, you need to add them to properties file as well, otherwise you will see a ParseException;

2. Those parameters are not really kafka parameters, they probably should be named: 'avro.schema.url' ...

src/main/java/com/pinterest/secor/io/impl/AvroFileReaderWriterFactory.java
```diff
@@ -0,0 +1,106 @@
...
+
+        public AvroFileWriter(LogFilePath path, CompressionCodec codec) throws IOException {
+
+            mWriter = new DataFileWriter(new GenericDatumWriter());
+
+            Path fsPath = new Path(path.getLogFilePath());
+            FileSystem fs = FileUtil.getFileSystem(path.getLogFilePath());
+            this.mCountingStream = new CountingOutputStream(fs.create(fsPath));
+
+            System.out.println(codec.getCompressorType().getCanonicalName());
```
HenryCaiHaiying: Remove System.out.println

src/main/scala/com/pinterest/secor/parser/AvroDateParser.scala
```diff
@@ -0,0 +1,76 @@
+package com.pinterest.secor.parser
```
HenryCaiHaiying: No scala code, please.

HenryCaiHaiying:
Thanks for the change, it's a lot cleaner now.  See my comments in the code.

And please don't add scala code, it's hard to maintain the scala binary code during scala version upgrade, even kafka is moving away from scala (the new consumer is re-written in Java).

ppearcy:
I've decided to change directions a little bit on this PR. The dependencies for avro and kafka schema repo add a decent amount of bloat and it's really not beneficial to have it part of core secor vs a standalone plugin (really, most of the parsers/writers should probably be independent). Also, there is a nasty bug in 1.0 Schema registry client that causes tons of load on schema registry, so need to use the 2.0-SNAPSHOT, which gets a little sketchy.  

Instead, these changes can be summarized as:
* New params for schema registry (I'd prefer to add support for dynamic params, so we don't need to change SecorConfig.java for plugin specific properties). 
* Add an Object to KeyValue to hold the parsed record. This allows the parsed record to flow through with the schema attached and avoids extra hops of reserializing it back out to write it to an avro container file. I'm not sold this will always be used for a parsedRecord, so maybe the naming could be tweaked, open to opinions. 

I'm planning on publishing the Avro registry secor plugin in my own repo.

ralph-tice:
@ppearcy Did this get published as a plugin? Can you drop a forward reference to it here for continuity?

HenryCaiHaiying:

PR #115, opened 2015-07-25
Add support to enable deploy/publish through SonaType to maven central

HenryCaiHaiying:
Added all the required change to pom to allow build publishing to maven central through SonaType.org.

Most of the changes are following examples in:
http://central.sonatype.org/pages/ossrh-guide.html

The first version is published to maven under (including both sources and javadoc jars):
https://repo1.maven.org/maven2/com/pinterest/secor/0.1

Since ver 0.1 is published, our current SNAPSHOT build will be upgraded to 0.2-SNAPSHOT

To publish the snapshot build (you need to have SonaType account and GnuPGP signing tools configured):
    # make sure the version in pom.xml ends with SNAPSHOT
    mvn clean deploy
    # the artifacts will stay in Sonatype repo (not going into maven central)

To publish a release build
    # make sure the version in pom.xml not ends with SNAPSHOT
    mvn clean deploy -P release
    # the artifacts are pushed all the way through to maven central
    # when the release build is published to maven central, make sure the version in pom.xml is upgraded to next-version-SNAPSHOT to reflect we are not building snapshot for next version until it's ready to be version released.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,11 +4,13 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
-    <groupId>com.pinterest.secor</groupId>
+    <groupId>com.pinterest</groupId>
     <artifactId>secor</artifactId>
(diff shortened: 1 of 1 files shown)


pgarbacki:
r+

PR #123, opened 2015-08-19
Support s3a hadoop file system.

kecaps:
Added tests to run through both with s3n and s3a
Updated hadoop-aws and jets3t packages.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -99,7 +99,7 @@
         <dependency>
             <groupId>net.java.dev.jets3t</groupId>
             <artifactId>jets3t</artifactId>
-            <version>0.7.1</version>
+            <version>0.9.4</version>
         </dependency>
         <dependency>
             <groupId>org.apache.kafka</groupId>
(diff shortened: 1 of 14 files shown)


pom.xml
```diff
@@ -99,7 +99,7 @@
         <dependency>
             <groupId>net.java.dev.jets3t</groupId>
             <artifactId>jets3t</artifactId>
-            <version>0.7.1</version>
+            <version>0.9.3</version>
```
pgarbacki: Other people identified issues with upgrading jets3t lib. You may want to follow https://groups.google.com/forum/#!topic/secor-users/6TlNaKuPLwA

pgarbacki:
@kecaps this PR broke the travis build. I'll have to revert it.

pgarbacki:
https://travis-ci.org/pinterest/secor/builds/78074163

kecaps:
The Travis ci testing has a timeout of 50 minutes. Since I doubled the
testing for both file systems it is exceeding the timeout. Would you prefer
removing test coverage or increasing the Travis timeout?

It may be fairly redundant to do all the tests for both file systems since
I think the s3 interaction of putting a file is the same in a lot of the
test scenarios.

pgarbacki:
@kecaps let's increase the timeout for now. We can remove the tests later when the runtime becomes unbearable.

kecaps:
I took a quick look at travis-ci documentation. It doesn't look like there
is a way to configure travis-ci to allow a build longer than 50 minutes:

http://docs.travis-ci.com/user/customizing-the-build/#Build-Timeouts
https://github.com/travis-ci/travis-ci/issues/2736

pgarbacki:
@kecaps let's remove the tests then

PR #125, opened 2015-08-25
Close consumers used in get(Committed|Last)Message.

liamstewart:
This stops connections from being opened and then immediately orphaned
on calls to these methods.


HenryCaiHaiying:
Looks good, thanks.

pgarbacki:
@liamstewart @HenryCaiHaiying the build started failing after merging this commit.

https://travis-ci.org/pinterest/secor

liamstewart:
I suspect a flaky test due to the fact that subsequent builds passed. Is the log file mentioned in the failing test available through travis?

PR #128, opened 2015-08-31
Google Cloud Storage integration

jgagnon1:
Allow to use Google Cloud Storage as storage backend instead of S3.

### DIFF
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -255,6 +255,18 @@ public class SecorConfig {
         return getString("secor.zookeeper.path");
     }
 
+    public String getGsCredentialsPath() {
+        return getString("secor.gs.credentials.path");
+    }
+
+    public String getGsBucket() {
+        return getString("secor.gs.bucket");
+    }
+
+    public String getGsPath() {
+        return getString("secor.gs.path");
+    }
+
     public boolean getBoolean(String name, boolean defaultValue) {
         return mProperties.getBoolean(name, defaultValue);
     }
--- a/src/main/java/com/pinterest/secor/uploader/GsUploadManager.java
+++ b/src/main/java/com/pinterest/secor/uploader/GsUploadManager.java
@@ -1,0 +1,108 @@
+package com.pinterest.secor.uploader;
+
+import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
+import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
+import com.google.api.client.http.FileContent;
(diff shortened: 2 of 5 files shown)


src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -255,6 +255,18 @@ public String getZookeeperPath() {
         return getString("secor.zookeeper.path");
     }
 
+    public String getGsCredentialsPath() {
+        return getString("secor.gs.credentials.path");
```
pgarbacki: Can you add those params to one of the .properties file and document them?

src/main/java/com/pinterest/secor/uploader/GsUploadManager.java
```diff
@@ -0,0 +1,108 @@
...
+     * globally shared instance across your application.
+     */
+    private static Storage mStorageService;
+
+    private Storage mClient;
+
+    public GsUploadManager(SecorConfig config) throws Exception {
+        super(config);
+
+        String credentialPath = System.getenv("GOOGLE_APPLICATION_CREDENTIALS");
```
pgarbacki: The priorities should be reversed: get the creds from the properties first and fall back to the env.

src/main/java/com/pinterest/secor/uploader/GsUploadManager.java
```diff
@@ -0,0 +1,108 @@
...
+                    throw new RuntimeException(e);
+                }
+            }
+        });
+
+        return new FutureHandle(f);
+    }
+
+    private static Storage getService(String credentialsPath) throws Exception {
+        if (null == mStorageService) {
```
pgarbacki: if (mStorageService == null)

pgarbacki:
@jgagnon1 Thanks for the contribution. Please address the comments and squash your commits.

src/main/java/com/pinterest/secor/common/SecorConfig.java
jgagnon1: Added in src/main/config/secor.dev.gs.partition.properties

jgagnon1:
@pgarbacki Done and done; squashed

jgagnon1:
Not sure to understand why the tests are failing after squashing?

jgagnon1:
I rebased, but not sure what caused tests exceed 50 minutes. @pgarbacki ??

pgarbacki:
@jgagnon1 :  I believe that this PR is at fault: https://github.com/pinterest/secor/pull/123

@kecaps are you looking into increasing the timeout?

kecaps:
I've taken out the s3a testing which was doubling the time of integration tests. If you resubmit this to travis-ci, it should no longer timeout.

PR #131, opened 2015-09-11
Lower wait time to speed up tests.

kecaps:
Looking at test configuration, I believe the max file age and kafka consumer timeout are both 10 seconds, not 1 minute.
I was able to lower the wait delay before verification from 120 seconds to 40 seconds. This sped up integration testing from 30 minutes to ~14 minutes.


HenryCaiHaiying:
I tried a few times on the travis build machine, it's all passing, so I merge this in.

PR #132, opened 2015-09-14
Fix DelimitedTextFile offset.

jgagnon1:
Rebased and merged with current master from #102 so it can be mergeable.

### DIFF
--- a/src/main/java/com/pinterest/secor/common/FileRegistry.java
+++ b/src/main/java/com/pinterest/secor/common/FileRegistry.java
@@ -17,6 +17,7 @@
 package com.pinterest.secor.common;
 
 import com.pinterest.secor.io.FileWriter;
+import com.pinterest.secor.io.impl.DelimitedTextFileReaderWriterFactory;
 import com.pinterest.secor.util.FileUtil;
 import com.pinterest.secor.util.ReflectionUtil;
 import com.pinterest.secor.util.StatsUtil;
@@ -38,12 +39,15 @@ public class FileRegistry {
     private static final Logger LOG = LoggerFactory.getLogger(FileRegistry.class);
 
     private final SecorConfig mConfig;
+    private final boolean mUseOffsetFile;
     private HashMap<TopicPartition, HashSet<LogFilePath>> mFiles;
     private HashMap<LogFilePath, FileWriter> mWriters;
     private HashMap<LogFilePath, Long> mCreationTimes;
 
-    public FileRegistry(SecorConfig mConfig) {
+    public FileRegistry(SecorConfig mConfig) throws ClassNotFoundException {
         this.mConfig = mConfig;
+        mUseOffsetFile = Class.forName(mConfig.getFileReaderWriterFactory())
+                == DelimitedTextFileReaderWriterFactory.class;
         mFiles = new HashMap<TopicPartition, HashSet<LogFilePath>>();
         mWriters = new HashMap<LogFilePath, FileWriter>();
         mCreationTimes = new HashMap<LogFilePath, Long>();
@@ -89,6 +93,10 @@ public class FileRegistry {
             // Just in case.
             FileUtil.delete(path.getLogFilePath());
             FileUtil.delete(path.getLogFileCrcPath());
+            if (mUseOffsetFile) {
+                FileUtil.delete(path.getLogFileOffsetPath());
+                FileUtil.delete(path.getLogFileOffsetCrcPath());
+            }
             TopicPartition topicPartition = new TopicPartition(path.getTopic(),
                     path.getKafkaPartition());
             HashSet<LogFilePath> files = mFiles.get(topicPartition);
@@ -127,6 +135,10 @@ public class FileRegistry {
         deleteWriter(path);
         FileUtil.delete(path.getLogFilePath());
         FileUtil.delete(path.getLogFileCrcPath());
+        if (mUseOffsetFile) {
+            FileUtil.delete(path.getLogFileOffsetPath());
+            FileUtil.delete(path.getLogFileOffsetCrcPath());
+        }
     }
 
     /**
--- a/src/main/java/com/pinterest/secor/common/LogFilePath.java
+++ b/src/main/java/com/pinterest/secor/common/LogFilePath.java
@@ -118,6 +118,10 @@ public class LogFilePath {
         return new LogFilePath(prefix, mTopic, mPartitions, mGeneration, mKafkaPartition, mOffset, mExtension);
     }
 
+    public LogFilePath offsetPath() {
+        return new LogFilePath(mPrefix, mTopic, mPartitions, mGeneration, mKafkaPartition, mOffset, ".offset");
+    }
+
     public String getLogFileParentDir() {
         ArrayList<String> elements = new ArrayList<String>();
         elements.add(mPrefix);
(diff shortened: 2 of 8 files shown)


src/main/scripts/run_tests.sh
```diff
@@ -407,25 +406,28 @@ start_s3
...
-        move_offset_back_test
-        if [ ${MESSAGE_TYPE} = "json" ]; then
-            post_and_verify_compressed_test
-        elif [ -z ${SKIP_COMPRESSED_BINARY} ]; then
-            post_and_verify_compressed_test
-        else
-            echo "Skipping compressed tests for ${MESSAGE_TYPE}"
-        fi
+       MESSAGE_TYPE=${key}
+       ADDITIONAL_OPTS=-Dsecor.file.reader.writer.factory=${READER_WRITERS[${key}]}
```
HenryCaiHaiying: How come -Dsecor.s3.filesystem argument is dropped?

src/main/scripts/run_tests.sh
```diff
@@ -407,25 +406,28 @@ start_s3
...
+       if [ ${MESSAGE_TYPE} = "json" ]; then
+           ADDITIONAL_OPTS="${ADDITIONAL_OPTS} -Dsecor.upload_offset_file=true"
```
HenryCaiHaiying: Do you need to test both upload_offset_file=true and false?

src/main/java/com/pinterest/secor/common/FileRegistry.java
```diff
@@ -38,12 +39,15 @@
...
     private final SecorConfig mConfig;
+    private final boolean mUseOffsetFile;
     private HashMap<TopicPartition, HashSet<LogFilePath>> mFiles;
     private HashMap<LogFilePath, FileWriter> mWriters;
     private HashMap<LogFilePath, Long> mCreationTimes;
 
-    public FileRegistry(SecorConfig mConfig) {
+    public FileRegistry(SecorConfig mConfig) throws ClassNotFoundException {
         this.mConfig = mConfig;
+        mUseOffsetFile = Class.forName(mConfig.getFileReaderWriterFactory())
```
HenryCaiHaiying: Why this check is not done through comparing against a config property?

src/main/java/com/pinterest/secor/common/LogFilePath.java
```diff
@@ -118,6 +118,10 @@ public LogFilePath withPrefix(String prefix) {
         return new LogFilePath(prefix, mTopic, mPartitions, mGeneration, mKafkaPartition, mOffset, mExtension);
     }
 
+    public LogFilePath offsetPath() {
```
HenryCaiHaiying: What's the difference between 'offsetPath' and 'getLogFileOffsetPath'?

src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -276,6 +276,8 @@ public boolean getBoolean(String name, boolean defaultValue) {
         return mProperties.getBoolean(name, defaultValue);
     }
 
+    public boolean getUploadOffsetFile() { return getBoolean("secor.upload_offset_file", false); }
```
HenryCaiHaiying: Probably the method name should be called 'useUploadOffsetFile', the parameter name probably should be called 'use_upload...' to indicate it's boolean type.

src/main/java/com/pinterest/secor/io/impl/DelimitedTextFileReaderWriterFactory.java
```diff
@@ -48,7 +49,7 @@
 
     @Override
     public FileReader BuildFileReader(LogFilePath logFilePath, CompressionCodec codec)
-            throws IllegalAccessException, IOException, InstantiationException {
+            throws Exception {
```
HenryCaiHaiying: We prefer individual exception classes.

src/main/java/com/pinterest/secor/io/impl/DelimitedTextFileReaderWriterFactory.java
```diff
@@ -58,76 +59,92 @@ public FileWriter BuildFileWriter(LogFilePath logFilePath, CompressionCodec code
...
+        private final SequenceFile.Reader mOffsetReader;
+        private final LongWritable mKey;
+        private final IntWritable mValueLength;
+        private final DataInputStream mReader;
         private Decompressor mDecompressor = null;
+        private byte[] mDelimiter;
 
-        public DelimitedTextFileReader(LogFilePath path, CompressionCodec codec) throws IOException {
+        public DelimitedTextFileReader(LogFilePath path, CompressionCodec codec) throws Exception {
+            Configuration config = new Configuration();
```
HenryCaiHaiying: This way you force the people to use offsetFile for DelimitedTextFileReader, this will break the existing DelimitedTextFileReader

HenryCaiHaiying:
High level comments, I didn't under your use case.  Why you need another offset file tracking records if you already have delimiter in the data file, why mix the offset file with the data file?

And you are requiring people use delimitedTextFile to use the offfsetFile, this will break their existing behaviors.

yyejun:
If you read original pull request, this change is to fix #101.

jgagnon1:
As @yyejun said; I just rebased #101 to make it mergeable. It seemed to fix an important issue that I could encounter since I use DelimitedTextFile. I didn't evaluate what the code was doing, so maybe @yyejun could answer your questions, otherwise I might take a look at it later on.

yyejun:
Current DelimitedTextFile class made a false assumption that the file name is always starting offset and entire file has continuous offset which was not true in a number of situations, e.g. double rebalance, date change. Current file trim method relies on correct offset, so it will not trim file correctly.

src/main/java/com/pinterest/secor/common/LogFilePath.java
yyejun: The 2nd method was added to accommodating a recent refactoring of upload manager.

src/main/scripts/run_tests.sh
yyejun: I added this flag purely for testing purpose.

src/main/java/com/pinterest/secor/common/FileRegistry.java
yyejun: Because this is not an optional fix for that class.

src/main/java/com/pinterest/secor/io/impl/DelimitedTextFileReaderWriterFactory.java
yyejun: Reader is only used for trimming purpose and secor does not perform any recovering of existing file. So I don't see how this can break any existing installation.

HenryCaiHaiying:
If we don't think DelimitedTextReaderWriter works without offset, we should have a new class: OffsettedTextReaderWriter which can use your double file writing.  Otherwise, why we still writes delimiters into the data file and we are not using that delimiter.

Once we start use double file writing, we need to be a bit more careful about the consistency between the files.  Maybe the data file writes always occur before offset file writes, and also it would be a good idea to add some sanity check between the two files (e.g. the sum of length file in offset file should be equal to the length of the data file).

yyejun:
@jgagnon1 Please feel free to make all necessary changes.

HenryCaiHaiying:

PR #133, opened 2015-09-16
GCS Direct upload

jgagnon1:
While running Google Cloud Storage in our test environment; I've come across some issue with the default uploading with Google SDK. Using the direct upload feature seems to have fixed some of the issues and using resumable upload is useless since we don't support that feature. I also added debug upload progress monitor in the doing.


HenryCaiHaiying:
Looks good, thanks.

PR #135, opened 2015-09-22
Updating Readme to match current Version in pom

damaex17:


HenryCaiHaiying:
Thanks for the change.  I didn't realize we have hard-coded secor version string in README file.  Can you do one step further, instead of hard-code another version in README file (which might change again in the near future), can you just use some version token string (e.g. secor-<latest-version>) in README file?

rud-bookbites:
Should probably be closed

HenryCaiHaiying:
But the version string way passed 0.2, I think we should just ask people to check the version string in pom.xml and reference that version in the README

rud-bookbites:
Yeah, this detail is in a bunch of places, a general "heads up" is probably more helpful

HenryCaiHaiying:
@rud-bookbites do you want that information in README file?

rud-bookbites:
There are multiple issues mentioning this problem. A bit of text in the readme would be helpful. Some more documentation would probably help even more, maybe in the form of guides or common use-cases. 

I suggested closing the issue as it is so old without activity is all.

HenryCaiHaiying:
Closed since the change was already out of date with the current code.

PR #138, opened 2015-09-26
Adding support for RegEx based parsing.

HenryCaiHaiying:
We have cases where the kafka message is a plain text and people use regex to extract the fields (e.g. timestamp).

Add support to allow people specify RegEx based parser, "message.timestamp.input.pattern" is used to specify a regex pattern to extract the timestamp, e.g.:

    ^[^ ]+ [^ ]+ ([^ ]+) .*$

The first group in the regex indicates the timestamp field

### DIFF
--- a/src/main/java/com/pinterest/secor/parser/RegexMessageParser.java
+++ b/src/main/java/com/pinterest/secor/parser/RegexMessageParser.java
@@ -1,0 +1,58 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.pinterest.secor.parser;
+
+import com.pinterest.secor.common.SecorConfig;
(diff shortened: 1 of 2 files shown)


src/main/java/com/pinterest/secor/parser/RegexMessageParser.java
```diff
@@ -0,0 +1,56 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * RegexMessageParser extracts timestamp field (specified by 'message.timestamp.input.pattern')
+ * The pattern specifies the regular exp to extract the timestamp field from a free-text line.
+ */
+public class RegexMessageParser extends TimestampedMessageParser {
+    private static final Logger LOG = LoggerFactory.getLogger(RegexMessageParser.class);
+
+    private final Pattern tsPattern;
```
pgarbacki: mTsPattern

src/main/java/com/pinterest/secor/parser/RegexMessageParser.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public RegexMessageParser(SecorConfig config) {
+        super(config);
+        String patStr = config.getMessageTimestampInputPattern();
+        LOG.info("timestamp pattern: " + patStr);
```
pgarbacki: From now on, lets standardize on the following logging format which avoids string concatenation overhead if logging is disabled:
LOG.info("timestamp pattern: {}", patStr);

src/main/java/com/pinterest/secor/parser/RegexMessageParser.java
```diff
@@ -0,0 +1,56 @@
...
+import com.pinterest.secor.message.Message;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * RegexMessageParser extracts timestamp field (specified by 'message.timestamp.input.pattern')
+ * The pattern specifies the regular exp to extract the timestamp field from a free-text line.
```
pgarbacki: author ...

pgarbacki:
@HenryCaiHaiying the code looks good but could you please squash the commits before merging?

PR #140, opened 2015-10-07
enable JsonMessageParser to parse ISO-8601 timestamps

flora:
@HenryCaiHaiying

### DIFF
--- a/src/test/java/com/pinterest/secor/parser/DateMessageParserTest.java
+++ b/src/test/java/com/pinterest/secor/parser/DateMessageParserTest.java
@@ -34,6 +34,7 @@ public class DateMessageParserTest extends TestCase {
     private Message mFormat1;
     private Message mFormat2;
     private Message mFormat3;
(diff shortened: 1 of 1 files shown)


src/main/java/com/pinterest/secor/parser/JsonMessageParser.java
```diff
@@ -36,7 +37,14 @@ public long extractTimestampMillis(final Message message) {
         if (jsonObject != null) {
             Object fieldValue = jsonObject.get(mConfig.getMessageTimestampName());
             if (fieldValue != null) {
-                return toMillis(Double.valueOf(fieldValue.toString()).longValue());
+                String timestamp = fieldValue.toString();
```
HenryCaiHaiying: This functionality (parsing date string in a DateFormatter) is already provided in DatesMessageParser.java

HenryCaiHaiying:

PR #141, opened 2015-10-12
Support user-defined upload policies

barata:
This PR adds support for user-defined upload policies. The upload policy must inherits abstract class UploadPolicy and implement 'shouldUpload(TopicPartition)' method. The subclass to be used as current policy is loaded from 'secor.uploadpolicy.strategy.class' property. Furthermore, an additional property 'secor.uploadpolicy.check.seconds' is used to determine how often the policy should be called.

Should the policy need to read any specific property, it can use 'secor.uploadpolicy.properties'. Properties are in the form key1=value1,key2=value2.

For example:

```
secor.uploadpolicy.check.seconds=600
secor.uploadpolicy.strategy.class=com.pinterest.secor.uploader.DefaultUploadPolicy
secor.uploadpolicy.properties=MAX_FILE_SIZE_BYTES=200000000,MAX_FILE_AGE_SECONDS=3600
```

In short, default values do not modify current Secor behavior. Therefore, DefaultUploadPolicy#shouldUpload() returns true if (size >= MAX_FILE_SIZE_BYTES OR modificationAgeSec >= MAX_FILE_AGE_SECONDS).

The following two properties were removed in order to use the new introduced model:

```
secor.max.file.size.bytes
secor.max.file.age.seconds
```

### DIFF
--- a/src/main/config/secor.dev.properties
+++ b/src/main/config/secor.dev.properties
@@ -34,8 +34,8 @@ kafka.seed.broker.port=9092
 zookeeper.quorum=localhost:2181
 
 # Upload policies.
-# 10K
-secor.max.file.size.bytes=10000
-# 10 seconds
-secor.max.file.age.seconds=10
-
+# Default upload policy
+secor.uploadpolicy.check.seconds=5
+secor.uploadpolicy.strategy.class=com.pinterest.secor.uploader.DefaultUploadPolicy
+# 10KB or 10 seconds (age of most recent file in a Secor partition)
+secor.uploadpolicy.properties=MAX_FILE_SIZE_BYTES=10000,MAX_FILE_AGE_SECONDS=10
--- a/src/main/config/secor.prod.properties
+++ b/src/main/config/secor.prod.properties
@@ -50,9 +50,9 @@ secor.swift.container=logsContainer
 ################
 
 # Upload policies.
-# 200MB
-secor.max.file.size.bytes=200000000
-# 1 hour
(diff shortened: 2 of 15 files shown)


barata:
@pgarbacki Could you please review my PR? It was motivated by some requirements from my company. We need a special policy that ensures events from Kafka do not take longer than X minutes to be uploaded to S3. That is why I also created a TimeBasedUploadPolicy.
Thanks a lot.

fabeschan:
secor.uploadpolicy.check.seconds is a useful addition. the current check period could result in creating vastly larger file sizes than specified in secor.max.file.size.bytes whenever the amount of data flow is large

barata:
@HenryCaiHaiying Do you think this is relevant? Can we proceed on this?

barata:
Thanks @alshamiri2 . Do you want me to rebase this PR? Do you have any plan to merge it?

HenryCaiHaiying:

barata:
Sync'ing is not a problem. I can do if you want. Just let me know if you find this PR useful.

HenryCaiHaiying:
There are quite a bit of change in the upload policy, take a look to see whether that satisfies your needs.  I don't quite like the idea of using this parameter format:
secor.uploadpolicy.properties=MAX_FILE_SIZE_BYTES=200000000,MAX_FILE_AGE_SECONDS=3600

The old way seems fine with me:
secor.max.file.size.bytes
secor.max.file.age.seconds

PR #142, opened 2015-10-14
Adding a shutdown hook to clean up the local files before exiting.

HenryCaiHaiying:
When secor process died, often we found there are a large set of local data files left on the disk.  Although we have a cleanup thread to clean files older than 24 hours, but sometimes this is already too late.

Add a shutdown hook to clean up files on process exit.

Changed a log message from debug to warning to help the debug.

### DIFF
--- a/src/main/java/com/pinterest/secor/common/FileRegistry.java
+++ b/src/main/java/com/pinterest/secor/common/FileRegistry.java
@@ -103,6 +103,11 @@ public class FileRegistry {
             mWriters.put(path, writer);
             mCreationTimes.put(path, System.currentTimeMillis() / 1000L);
             LOG.debug("created writer for path {}", path.getLogFilePath());
+            LOG.debug("Register deleteOnExit for path {}", path.getLogFilePath());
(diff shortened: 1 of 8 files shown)


pgarbacki:
Shutdown hooks are fragile. I would prefer us to use File.deleteOnExit() instead. You could set it up at the directory level.

src/main/java/com/pinterest/secor/common/FileRegistry.java
```diff
@@ -103,6 +103,9 @@ public FileWriter getOrCreateWriter(LogFilePath path, CompressionCodec codec)
             mWriters.put(path, writer);
             mCreationTimes.put(path, System.currentTimeMillis() / 1000L);
             LOG.debug("created writer for path {}", path.getLogFilePath());
+            LOG.debug("Register deleteOnExit for path {}", path.getLogFilePath());
+            FileUtil.deleteOnExit(path.getLogFilePath());
```
pgarbacki: Why do we do it at the individual file level rather than dir level? Each consumer is writing to a dir composed of PID and thread id, so those dirs are specific to a given instance.

PR #145, opened 2015-11-18
Backward compatibility fix and GS compatibility fix

jgagnon1:
The introduction of cloud.service was being to restrictive when adding new kind of service (IE: Google cloud storage) always fallbacking to S3 for validation and such. I made it optional (but more validation when specifying it) so it could be more flexible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Build Status](https://travis-ci.org/pinterest/secor.svg)](https://travis-ci.org/pinterest/secor)
 
-Secor is a service persisting [Kafka] logs to [Amazon S3] and [Openstack Swift].
+Secor is a service persisting [Kafka] logs to [Amazon S3], [Google Cloud Storage] and [Openstack Swift].
 
 ## Key features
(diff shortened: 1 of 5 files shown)


src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -209,7 +209,7 @@ public int getOstrichPort() {
     }
 
     public String getCloudService() {
-        return getString("cloud.service");
+        return getString("cloud.service", "");
```
HenryCaiHaiying: why change this line?  The original default value is 'S3' which is specified in the properties file.  I think we always want people to explicitly set the properties value in the properties file, not in the java code.

jgagnon1: It felt useless to me; but you are right this is not necessary to put this default value; I'll change it

jgagnon1:
Retriggering build; Circle CI failed randomly ^

HenryCaiHaiying:
Looks good, thanks.

PR #147, opened 2015-11-23
Archive kafka message key.

### LINKED ISSUE #121: Support archiving of Kafka tombstones
A kafka tombstone is used with topics having hash keys to indicate data for the key has been removed from the log stream.  The log cleaner/compacter can be configured to eventually remove these tombstones.  Tombstones are kafka log messages with a null payload.

Currently, it appears that Secor has no support for tracking tombstones through to the archive, as it does not even retain the key of the message when reading from Kafka.  It will simply write out the offset of the tombstone, with no information about what was actually deleted (assuming it doesn't run into one of the other reported bugs about null payloads). This information is critical if the downstream consumers are to make sense of deletes in the archived stream.

I believe that this may require an output file format change, as a triple of key/offset/payload must be stored.

llamahunter:
Create a new SequenceFile Reader/Writer that stores a MessagePack in the key
which contains both the offset and the kafka message key.  A number of small
API changes (primarily to Message and KeyValue) to plumb the kafka message
key through from the consumer to the writer.  Added a new unit test for the
new Reader/Writer.

fixes #121


llamahunter:
This is my initial attempt at archiving kafka message keys with Secor.  Please let me know if something is amiss with this approach and I'll fix it and resubmit the pull request.

As you can see, I decided to leave the existing SequenceFileReaderWriter alone and simply make a new one, as the Key field was changing from Long to Bytes.  There's no real config file changes, as one just needs to specify the different secor.file.reader.writer.factory to get the MessagePack version.

llamahunter:
Not sure what's up with the travis-ci failure on openjdk7.  I didn't change the pom.xml file, and added no new build time dependencies that I know of.  The oraclejdk7 build succeeded.  Is there a known problem with the openjdk7 build?

HenryCaiHaiying:
I manually kicked off another run of the integration test to see if the
error is just transient.

src/main/java/com/pinterest/secor/io/impl/MessagePackSequenceFileReaderWriterFactory.java
```diff
@@ -0,0 +1,165 @@
...
+            FileSystem fs = FileUtil.getFileSystem(path.getLogFilePath());
+            this.mReader = new SequenceFile.Reader(fs, fsPath, config);
+            this.mKey = (BytesWritable) mReader.getKeyClass().newInstance();
+            this.mValue = (BytesWritable) mReader.getValueClass().newInstance();
+        }
+
+        @Override
+        public KeyValue next() throws IOException {
+            if (mReader.next(mKey, mValue)) {
+                MessageUnpacker unpacker = MessagePack.newDefaultUnpacker(mKey.getBytes());
```
HenryCaiHaiying: Is 'unpacker' object expensive to create?  Should the object creation be done in the constructor?

src/main/java/com/pinterest/secor/io/impl/MessagePackSequenceFileReaderWriterFactory.java
```diff
@@ -0,0 +1,165 @@
...
+                            break;
+                        case KAFKA_HASH_KEY:
+                            int keySize = unpacker.unpackBinaryHeader();
+                            keyBytes = new byte[keySize];
+                            unpacker.readPayload(keyBytes);
+                            break;
+                    }
+                }
+                unpacker.close();
+                return new KeyValue(offset, keyBytes, Arrays.copyOfRange(mValue.getBytes(), 0, mValue.getLength()));
```
HenryCaiHaiying: Should we check that offset is present in the messagePack?

src/main/java/com/pinterest/secor/message/Message.java
```diff
@@ -44,13 +48,17 @@ public String toString() {
...
     }
 
-    public Message(String topic, int kafkaPartition, long offset, byte[] payload) {
+    public Message(String topic, int kafkaPartition, long offset, byte[] key, byte[] payload) {
         mTopic = topic;
         mKafkaPartition = kafkaPartition;
         mOffset = offset;
+        mKey = key;
+        if (mKey == null) {
+            mKey = EMPTY_BYTES;
```
HenryCaiHaiying: When will the key be null?  For tombstone, the payload is null.

src/main/java/com/pinterest/secor/io/impl/MessagePackSequenceFileReaderWriterFactory.java
llamahunter: They are sort of designed to be single use objects. It doesn't have a zero argument constructor... must be constructed with a specific input stream.  There is a reset() call for changing streams, but don't see much use of that in example code.

src/main/java/com/pinterest/secor/io/KeyValue.java
```diff
@@ -25,16 +25,29 @@
  */
 public class KeyValue {
 	
-	private final long mKey;
+	private final long mOffset;
+	private final byte[] mKey;
```
HenryCaiHaiying: I think we should renamed the old 'mKey' to 'mKafkaKey' (so getKey() would become getKafkaKey()) this way we wouldn't run into the danger of confusing the usage of the 'getKey()' in one of the caller site.

src/test/java/com/pinterest/secor/parser/DateMessageParserTest.java
```diff
@@ -44,19 +44,19 @@ public void setUp() throws Exception {
...
                 .getBytes("UTF-8");
-        mFormat1 = new Message("test", 0, 0, format1);
+        mFormat1 = new Message("test", 0, 0, null, format1);
```
HenryCaiHaiying: Should we give a KafkaKey to those messages?

src/main/java/com/pinterest/secor/io/impl/MessagePackSequenceFileReaderWriterFactory.java
llamahunter: It's initialized to 0.  Not sure what I would do otherwise if it's not there.

src/main/java/com/pinterest/secor/message/Message.java
llamahunter: Keys are optional in kafka.  They can be null if the user wants a random distribution across partitions.  Of course, log compaction and tombstones makes no sense without keys.

src/main/java/com/pinterest/secor/io/KeyValue.java
llamahunter: so there would be getKey() which returns the offset, and getKafkaKey() which returns the key?  That seems more confusing.

src/test/java/com/pinterest/secor/parser/DateMessageParserTest.java
llamahunter: No need for these tests, which are about payload parsing.

src/main/java/com/pinterest/secor/io/KeyValue.java
HenryCaiHaiying: There is only going to be getKafkaKey() and getOffset() methods in the new class.  This would force all the client to rename the invocation of getKey() to getOffset().  I was worried if we reusing the getKey() with different meanings, some of the client might forget to rename 'getKey()' to 'getOffset'.

llamahunter: Ah.. got it.  Will make that change.

llamahunter:
Unclear what failed this time.  These tests seem fragile.

HenryCaiHaiying:
I rerun the integration tests, they are OK now.  Richard, are you going to rename the method getKey() on Message class to getKafkaKey()?

llamahunter:
Already changed getKey to getKafkaKey() about 10 hours ago.  Do you not see the new diffs?  Hmm...

HenryCaiHaiying:
Is that in Message.java?  I don't see the getKafkaKey in the PR.

llamahunter:
You want Message.java changed as well?  It didn't previously have a getKey() method.

llamahunter:
Alright.  Message and ParsedMessage also updated.

HenryCaiHaiying:
Looks good.  I will wait to see if the Travis CI can pass this time and then merge in your change.

llamahunter:
Do you have a 'contributor agreement' that needs to be signed by my employer, or do you use the 'contribution' definition in the apache license?

HenryCaiHaiying:
I think we are just using the standard contribution agreement from apache license.  Is there anything specific you are referring to?

@pgarbacki to see if he has more insights.

llamahunter:
Our lawyers are somewhat pedantic/paranoid about opensource software.  If you don't have a contributor agreement, that's probably fine, as the general license is Apache, and it defines the scope of patent grants and litigation, and warranties around contributions.  Our lawyers just wanted to know if there was any separate agreement.

llamahunter:
For reasons... see https://en.wikipedia.org/wiki/Tivoization
GPLv3 was motivated largely by Stallman's vendetta against my employer.

HenryCaiHaiying:
I am not the expert in those licenses.  We are using the standard apache
license: https://github.com/pinterest/secor/blob/master/LICENSE

Let me know if you still want us to merge your PR.

llamahunter:
Yup, please do.  Just needed to know if you had any additional agreements that needed to be signed.

HenryCaiHaiying:
Richard, I merged in your PR.  I am not aware of any extra agreements that needs to be signed.

llamahunter:
woot! thx!

PR #149, opened 2015-12-04
Use the strict UNIX_LINES mode to do regular expression parsing

HenryCaiHaiying:
Recently we added RegexMessageParser to parse the text log file, occasionally I saw the error message about cannot parsing some messages.

After investigation, I realized that the problem is some text log contain non-standard unicode (e.g. 0x85) in the url field, these 0x85 char are being treated as newline in java.text.Pattern.

Fix this by using strict UNIX line parsing (i.e. only recognize 0x0a as the line terminator)


pgarbacki:
r+

PR #156, opened 2015-12-15
Added options for S3 managed and customer-key based encryption, fixed…

keithdotpower:
… dependency issues arising from using newer AWS S3 SDK version

### DIFF
--- a/src/main/java/com/pinterest/secor/uploader/S3UploadManager.java
+++ b/src/main/java/com/pinterest/secor/uploader/S3UploadManager.java
@@ -16,9 +16,12 @@
  */
 package com.pinterest.secor.uploader;
 
+import com.amazonaws.services.s3.model.ObjectMetadata;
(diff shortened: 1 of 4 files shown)


src/main/java/com/pinterest/secor/uploader/S3UploadManager.java
```diff
@@ -41,10 +44,15 @@
  * default credential provider chain (supports environment variables,
  * system properties, credientials file, and IAM credentials).
  *
+ * If set, it will
```
jgagnon1: Incomplete comment here.

keithdotpower: I'll have to open another PR I'm afraid. I had to make my fork private to integrate it with our build system and only realised I couldn't continue with this PR afterwards. I've made a new fork but there appears to be no way to link these up. Will close this now.

PR #157, opened 2015-12-16
Add a new config to allow more flexible mapping from kafka topic to h…

HenryCaiHaiying:
…ive table name

Currently, we have secor.hive.prefix to allow people adding prefix to a kakfa topic name when mapping to hive table name.  Sometimes, people have an existing hive table name which doesn't fall in into the prefix pattern.  Add new config 'secor.hive.table.name.topic1=table1' to map topic name to arbituary hive table name.  This is an optional config, the existence of this property will override the hive.table.prefix property.

### DIFF
--- a/src/main/config/secor.common.properties
+++ b/src/main/config/secor.common.properties
@@ -166,10 +166,16 @@ secor.local.log.delete.age.hours=-1
 # It is available at https://api.qubole.com/users/edit
 qubole.api.token=
 
-# hive tables are generally named after the topics. For instance if the topic is request_log
-# the hive table is also called request_log. If you want this to be pinlog_request_log you can
-# set this config to "pinlog_". This affects all topics.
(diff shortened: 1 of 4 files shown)


src/main/config/secor.common.properties
```diff
@@ -171,6 +171,11 @@ qubole.api.token=
 # set this config to "pinlog_". This affects all topics.
 hive.table.prefix=
 
+# You can also name your hive table directly if your hive table doesn't 
+# follow the pattern of <hive.table.prefix><kafka topic>
+# E.g.  hive.table.name.topic1=table1 to indicate that hive table for 
+# kafka topic <topic1> will be named <table1>
```
pgarbacki: We should probably have a separate config file for finalizer-specific params.

pgarbacki:
Tests are failing. Other than that, r+

PR #159, opened 2015-12-16
Support for S3 encryption

keithdotpower:


src/main/config/secor.common.properties
```diff
@@ -46,6 +46,30 @@ aws.secret.key=
...
+###########################
+# START AWS S3 ENCRYPTION #
+###########################
+
+# AWS specify type of server-side encryption, if any
+# set to S3 to enable S3-managed encryption
+# set to KMS to enable AWS KMS-managed encryption (see aws.sse.kms.key)
+# set to customer to enable customer-managed encryption (see aws.sse.customer.key)
+# set empty to disable encryption
+aws.sse.type=
```
HenryCaiHaiying: If you don't set the value for this new property, does it throw exception when you call mConfig.getProperty on this new one?

keithdotpower: Hi Henry,

Thanks for the review.

The intent is that if that property is not set, it should just not use
encryption. That's what the last line in the comment is supposed to mean.
But I don't think that's the part you're talking about.

I just took a look at the code that reads this property in S3UploadManager.
There's a null check, but I'm pretty sure it should be checking isEmpty()
instead. So currently it looks like that code that checks for valid values
will throw an IllegalArgumentException if the property is left unspecified.
I guess this is what you're anticipating. I'm pretty sure I already noticed
and fixed this locally but with the various copying and rebasing I've been
doing to get his into shape for this pull request I must've missed it.

I can easily fix it but I'll hold off and find a way to add tests. I had a
look at the existing tests and it didn't look like the S3 uploading code
was being tested but I'll at least find a way to test the new config
switches.

It'll be next week before I can get to it though.

Cheers,

Keith


On 16 December 2015 at 20:12, Henry Haiying Cai <notifications@github.com>
wrote:

> In src/main/config/secor.common.properties
> <https://github.com/pinterest/secor/pull/159#discussion_r47825371>:
>
> > @@ -46,6 +46,30 @@ aws.secret.key=
> >  aws.region=
> >  aws.endpoint=
> >
> > +###########################
> > +# START AWS S3 ENCRYPTION #
> > +###########################
> > +
> > +# AWS specify type of server-side encryption, if any
> > +# set to S3 to enable S3-managed encryption
> > +# set to KMS to enable AWS KMS-managed encryption (see aws.sse.kms.key)
> > +# set to customer to enable customer-managed encryption (see aws.sse.customer.key)
> > +# set empty to disable encryption
> > +aws.sse.type=
>
> If you don't set the value for this new property, does it throw exception
> when you call mConfig.getProperty on this new one?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pinterest/secor/pull/159/files#r47825371>.
>



-- 

*Keith Power, PhD *- Data Science Engineer




Your operations product, for security.
Mobile: +353 86 2028099
Skype: keithpower
Twitter: @keithpippower, @barricadeio

HenryCaiHaiying: Keith,

I think you can do a quick testing in your java code to see what the value
for this new property in java code, (whether it's null or empty string or
get an exception) if you just put 'property=' in the properties file.  As
long as it doesn't throw exception, we should be fine.

keithdotpower:
Henry,

It works fine now if the property is unspecified.

cheers,

Keith

HenryCaiHaiying:
Thanks for the work.

keithdotpower:
Thank you for yours.

PR #162, opened 2016-01-04
Add support for timestamps in fields beyond the first to Thrift date parser

jdeastwood:
This PR adds support for parsing dates from thrift messages from fields other than the first via a new parameter `message.timestamp.id`. For simplicity the field still needs to be at the top level of the thrift object. 

Also added a `message.timestamp.type` parameter which I've intentionally not mentioned in the documentation update -- if folks want to subject themselves to the 2038 problem like us, they'l have to dig for it.

If these config values are unset the behaviour defaults to the old behaviour of using the first field.

### DIFF
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -296,6 +296,14 @@ public class SecorConfig {
         return getString("message.timestamp.name");
     }
 
+    public int getMessageTimestampId() {
+        return getInt("message.timestamp.id");
+    }
+
+    public String getMessageTimestampType() {
+        return getString("message.timestamp.type");
+    }
+
     public String getMessageTimestampInputPattern() {
         return getString("message.timestamp.input.pattern");
     }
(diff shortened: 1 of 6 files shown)


src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -296,6 +296,14 @@ public String getMessageTimestampName() {
         return getString("message.timestamp.name");
     }
 
+    public int getMessageTimestampId() {
+        return getInt("message.timestamp.id", 1);
+    }
+
+    public String getMessageTimestampType(String defaultValue) {
```
HenryCaiHaiying: Can the defaultValue specified in the property file?  So far we have keep the convention to have the default value explicitly spelled out in the property file.

HenryCaiHaiying:
Thanks for the work.  A few general comments:

1. Have the default value specified in the property file explicitly;
2. Can you squash 5 commits into on 1?
3. Once the timestamp field is not field 1 anymore, there is a performance overhead since the deserializer needs to read through the previous n-1 fields until it hits your Nth field.

jdeastwood:
Hi @HenryCaiHaiying 

> Have the default value specified in the property file explicitly;
Done!

> Can you squash 5 commits into on 1?
Done!

> Once the timestamp field is not field 1 anymore, there is a performance overhead since the deserializer needs to read through the previous n-1 fields until it hits your Nth field.
Yes :(. I've now added a warning to that effect next to the config parameter. 

FWIW I ran some simple benchmarks: https://gist.github.com/jdeastwood/114226b4127975b2034e (measuring throughput -- messages deserialized / s)

Looks like there is a performance penalty for reading past the first field regardless of the size of the message (although I'm a bit suspicious of that particular component of the benchmark, since is it the same message each time...). However its not so huge that it isn't useful when sticking close to the start of the message (our use case for example is just the third field).

HenryCaiHaiying:
Looks good, thanks.

PR #163, opened 2016-01-06
Kafka : add kafka consumer properties to allow offset migration

nguyenvanthan:
Hi,

With this pull request, you should be able to migrate offset from zookeeper to kafka consumer topic.
Then we can use Secor with kafka 0.9.0.  
[Kafka consumer documentation](http://kafka.apache.org/documentation.html#oldconsumerconfigs)

In the second commit, i just add my company to Secor users. We use it for several months in production.

I'm open to any suggestions :smile:

### DIFF
--- a/src/main/config/secor.common.properties
+++ b/src/main/config/secor.common.properties
@@ -139,6 +139,16 @@ kafka.seed.broker.port=9092
 # to as the chroot.
 kafka.zookeeper.path=/
 
+# Store offset in zookeeper and kafka consumer topic.
+# Only used if kafka.offsets.storage is set to "kafka"
+# http://kafka.apache.org/documentation.html#oldconsumerconfigs
+# Possible values: true or false
+kafka.dual.commit.enabled=true
+
+# Storage offset.
+# Possible values: "zookeeper" to read offset from zookeeper or "kafka" to read offset from kafka consumer topic
+kafka.offsets.storage=zookeeper
+
 # Secor generation is a version that should be incremented during non-backwards-compabile
 # Secor releases. Generation number is one of the components of generated log file names.
 # Generation number makes sure that outputs of different Secor versions are isolated.
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -117,6 +117,14 @@ public class SecorConfig {
         return getString("kafka.fetch.wait.max.ms");
     }
 
+    public String getDualCommitEnabled() {
+        return getString("kafka.dual.commit.enabled");
+    }
+
+    public String getOffsetsStorage() {
+        return getString("kafka.offsets.storage");
+    }
+
     public int getGeneration() {
         return getInt("secor.generation");
     }
--- a/README.md
+++ b/README.md
@@ -131,6 +131,7 @@ Secor is distributed under [Apache License, Version 2.0](http://www.apache.org/l
   * [Yelp](http://www.yelp.com)
   * [VarageSale](http://www.varagesale.com)
   * [Skyscanner](http://www.skyscanner.net)
+  * [Nextperf](http://www.nextperf.com)
 
 ## Help
 
--- a/pom.xml
+++ b/pom.xml
@@ -277,6 +277,11 @@
                 <directory>src/main/config</directory>
             </resource>
         </resources>
+        <testResources>
+            <testResource>
+                <directory>src/test/config</directory>
+            </testResource>
+        </testResources>
         <plugins>
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
--- a/src/main/java/com/pinterest/secor/reader/MessageReader.java
+++ b/src/main/java/com/pinterest/secor/reader/MessageReader.java
@@ -115,6 +115,10 @@ public class MessageReader {
         props.put("auto.offset.reset", "smallest");
         props.put("consumer.timeout.ms", Integer.toString(mConfig.getConsumerTimeoutMs()));
         props.put("consumer.id", IdUtil.getConsumerId());
+        // Properties required to upgrade from kafka 0.8.x to 0.9.x
+        props.put("dual.commit.enabled", mConfig.getDualCommitEnabled());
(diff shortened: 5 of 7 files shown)


src/main/config/secor.common.properties
```diff
@@ -139,6 +139,14 @@ kafka.seed.broker.port=9092
 # to as the chroot.
 kafka.zookeeper.path=/
 
+# Store offset in zookeeper and kafka consumer topic.
+# Possible values: true or false
+kafka.dual.commit.enabled=true
```
HenryCaiHaiying: Also indicate that this parameter is only used when offsets.storage is set to 'kafka', and give a link to kafka configuration doc.

src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -115,6 +116,14 @@ public String getFetchWaitMaxMs() {
         return getString("kafka.fetch.wait.max.ms");
     }
 
+    public boolean getDualCommitEnabled() {
+        return Boolean.parseBoolean(getStringLastValue("kafka.dual.commit.enabled", "true"));
```
HenryCaiHaiying: We prefer the default specified in the properties file.

src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -115,6 +116,14 @@ public String getFetchWaitMaxMs() {
...
+    }
+
+    public String getOffsetsStorage() {
+        return getStringLastValue("kafka.offsets.storage", "zookeeper");
```
HenryCaiHaiying: default specified in property file.

src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -381,6 +390,11 @@ private String getString(String name) {
         return mProperties.getString(name);
     }
 
+    private String getString(String name, String defaultValue) {
```
HenryCaiHaiying: No need for this new method.

src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -397,4 +411,19 @@ private long getLong(String name) {
     private String[] getStringArray(String name) {
         return mProperties.getStringArray(name);
     }
+
+    // This method can be useful if you use include and have several times the same key
+    // Retrieve the last children file value
+    @VisibleForTesting
+    String getStringLastValue(String name, String defaultValue) {
```
HenryCaiHaiying: Do you really this method?  This behavior is different than all the other property value which we got from the first occurrence.

HenryCaiHaiying:
Good work.  A few generate comments:

- Please squash your multiple commits into one;
- We usually prefer the default value specified in property file
- Doc that 'dual.commit.enabled' is only taken into effect when the storage is kafka, give a link to the kakfa doc;
- getStringLast is dubious, it changed the behavior of how we treat other properties.

src/main/config/secor.common.properties
```diff
@@ -139,6 +139,14 @@ kafka.seed.broker.port=9092
...
+
+# Storage offset.
+# Possible values: "zookeeper" to read offset from zookeeper or "kafka" to read offset from kafka consumer topic
+kafka.offsets.storage=zookeeper
+
```
HenryCaiHaiying: Maybe we should also add 'offset.storage.topic' property for people to specify the topic name to store the offset.

pgarbacki:
@nguyenvanthan when you are done with your changes, please squash the commits.

src/main/java/com/pinterest/secor/common/SecorConfig.java
nguyenvanthan: I don't find any other method to override properties.
When we include a file into another, i want only the child properties but the default behaviour is to return the first value (the parent, included file).

src/main/java/com/pinterest/secor/common/SecorConfig.java
nguyenvanthan: Ok i will remove this

src/main/config/secor.common.properties
nguyenvanthan: I don't see this property for consumer, i only seen it in kafka connect part.
Are you sure that it will define the consumer offset storage topic ?

nguyenvanthan:
Hi,
i just need an advice on the properties part.
On several framework, tools, when we have inheritance or inclusion, it is more natural to have the "child" value (last defined one) but with PropertiesConfiguration, we have the "parent" value.

I work for several months with secor and i use different files that define different "jobs" to copy specific topic in specific S3 bucket. Then when i want to modify a key that exist in "common" files, i have to remove the key then to duplicate them into "jobs" files.
I think, it would be more convenient if i could easily override the key in my "child" file.

Do you want to keep the parent value for every properties ?
Or maybe, can I just make ```getStringLastValue``` method public and use it in ```MessageReader```.

nguyenvanthan:
@pgarbacki @HenryCaiHaiying i push another commit. If you are agree with it i will squash the commits.

src/main/config/secor.common.properties
HenryCaiHaiying: I am not sure about that either, if you got some time, you can give it a try for that parameter, otherwise you can ignore my comment.

HenryCaiHaiying:
For the property overriding order, I don't have a good suggestion either.  I like the idea of taking the last occurrence of that property, but that will break the current behavior.  And we don't want this property be handled differently comparing to the the other properties.  @pgarbacki  to see his comments.

Two solutions I can think of, 
1. adding a property override control parameter, this parameter decides whether we take the first occurrence or last occurrence of the property, this will apply for all properties.
2. Is that possible to have a tool in your env to scan/update your property file and make the last one wins.

pgarbacki:
@HenryCaiHaiying I like suggestion 2. Overwriting properties is confusing.

nguyenvanthan:
@HenryCaiHaiying @pgarbacki I remove the properties override part, i only rebase and commit the new 2 properties.

HenryCaiHaiying:
Looks good, thanks for the contribution

PR #164, opened 2016-01-08
Logic to read timezone from config

sgmarghade:
TimestampedMessageParser has hardcoded UTC timezone. This change will read property from config file with default value as UTC which is handled in TimestampedMessageParser. 

This will help in creating partitions in s3 as per timezone provided.

### DIFF
--- a/src/main/java/com/pinterest/secor/parser/TimestampedMessageParser.java
+++ b/src/main/java/com/pinterest/secor/parser/TimestampedMessageParser.java
@@ -54,11 +54,12 @@ public abstract class TimestampedMessageParser extends MessageParser implements 
         LOG.info("FinalizerDelaySeconds: {}", mFinalizerDelaySeconds);
 
         mDtFormatter = new SimpleDateFormat("yyyy-MM-dd");
-        mDtFormatter.setTimeZone(TimeZone.getTimeZone("UTC"));
+        TimeZone timeZone = config.getTimeZone() != null ? TimeZone.getTimeZone(config.getTimeZone()) : TimeZone.getTimeZone("UTC");
+        mDtFormatter.setTimeZone(timeZone);
         mHrFormatter = new SimpleDateFormat("HH");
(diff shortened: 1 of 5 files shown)


src/main/java/com/pinterest/secor/parser/TimestampedMessageParser.java
```diff
@@ -54,11 +54,12 @@ public TimestampedMessageParser(SecorConfig config) {
         LOG.info("FinalizerDelaySeconds: {}", mFinalizerDelaySeconds);
 
         mDtFormatter = new SimpleDateFormat("yyyy-MM-dd");
-        mDtFormatter.setTimeZone(TimeZone.getTimeZone("UTC"));
+        TimeZone timeZone = config.getTimeZone() != null ? TimeZone.getTimeZone(config.getTimeZone()) : TimeZone.getTimeZone("UTC");
```
HenryCaiHaiying: Can you do null or empty string check instead?  And if it's an invalid timezone string, do you want to fall back to UTC?

src/main/java/com/pinterest/secor/tools/LogFileDeleter.java
```diff
@@ -44,7 +44,8 @@ public void deleteOldLogs() throws Exception {
         }
         String[] consumerDirs = FileUtil.list(mConfig.getLocalPath());
         SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss z");
-        format.setTimeZone(TimeZone.getTimeZone("UTC"));
+        TimeZone timeZone = mConfig.getTimeZone() != null ? TimeZone.getTimeZone(mConfig.getTimeZone()) : TimeZone.getTimeZone("UTC");
```
HenryCaiHaiying: Looks like this piece of code is also repeated here, probably better to move that into a helper method.

HenryCaiHaiying:
Thanks for the change.  When you are done, please also squash your commits into one.

PR #165, opened 2016-01-10
Logic to read timezone from config

sgmarghade:
Accidentally deleted old merge request. 
1. Timezone reading has moved to config file. 
2. as per Timezone api doc,

      : TimeZone.getTimeZone(<id> )  will return the specified TimeZone, or the GMT zone if the given ID cannot be understood.

### DIFF
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -367,8 +367,9 @@ public class SecorConfig {
         return getInt("partitioner.finalizer.delay.seconds");
     }
 
-    public TimeZone getTimeZone(){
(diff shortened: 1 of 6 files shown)


src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -365,6 +367,10 @@ public int getFinalizerDelaySeconds() {
         return getInt("partitioner.finalizer.delay.seconds");
     }
 
+    public TimeZone getTimeZone(){
+        return Strings.isNullOrEmpty(getString("secor.parser.timezone")) ?  TimeZone.getTimeZone("UTC") : TimeZone.getTimeZone(getString("secor.parser.timezone"));
```
HenryCaiHaiying: This line is a bit too long, can you break it into two?  I think most of the coding style has 100 chars line width.

sgmarghade:
Changes pushed as suggested.

HenryCaiHaiying:
Looks good, thanks.

PR #167, opened 2016-01-12
Add support for getting timestamp field out of nested structures

mcehlert:
Hi,
with this PR support for timestamp fields in nested JSON structures is added.
This affects JsonMessageParser and DateMessageParser.

### DIFF
--- a/src/main/java/com/pinterest/secor/parser/DateMessageParser.java
+++ b/src/main/java/com/pinterest/secor/parser/DateMessageParser.java
@@ -52,7 +52,7 @@ public class DateMessageParser extends MessageParser {
         String result[] = { defaultDate };
 
         if (jsonObject != null) {
-            Object fieldValue = jsonObject.get(mConfig.getMessageTimestampName());
+            Object fieldValue = getJsonFieldValue(jsonObject);
             Object inputPattern = mConfig.getMessageTimestampInputPattern();
             if (fieldValue != null && inputPattern != null) {
                 try {
--- a/src/main/java/com/pinterest/secor/parser/MessageParser.java
+++ b/src/main/java/com/pinterest/secor/parser/MessageParser.java
@@ -19,6 +19,11 @@ package com.pinterest.secor.parser;
 import com.pinterest.secor.common.SecorConfig;
 import com.pinterest.secor.message.Message;
 import com.pinterest.secor.message.ParsedMessage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import net.minidev.json.JSONObject;
+import net.minidev.json.JSONValue;
+import java.util.regex.Pattern;
 
 // TODO(pawel): should we offer a multi-message parser capable of parsing multiple types of
 // messages?  E.g., it could be implemented as a composite trying out different parsers and using
@@ -31,9 +36,15 @@ import com.pinterest.secor.message.ParsedMessage;
  */
 public abstract class MessageParser {
     protected SecorConfig mConfig;
+    protected String[] mNestedFields;
+    private static final Logger LOG = LoggerFactory.getLogger(MessageParser.class);
 
     public MessageParser(SecorConfig config) {
         mConfig = config;
+        if (mConfig.getMessageTimestampName() != null && mConfig.getMessageTimestampNameSeparator() != null) {
+            String separatorPattern = Pattern.quote(mConfig.getMessageTimestampNameSeparator());
+            mNestedFields = mConfig.getMessageTimestampName().split(separatorPattern);
+        }
     }
 
     public ParsedMessage parse(Message message) throws Exception {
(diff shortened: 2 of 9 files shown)


jgagnon1:
Simple question; is there any reason why NOT to always use . (dot) as separator ? I feel like this is kind of useless to leave the choice to the user to decide the separator for the configuration.

src/main/java/com/pinterest/secor/parser/DateMessageParser.java
```diff
@@ -52,7 +52,27 @@ public DateMessageParser(SecorConfig config) {
         String result[] = { defaultDate };
 
         if (jsonObject != null) {
-            Object fieldValue = jsonObject.get(mConfig.getMessageTimestampName());
+        	String messageTimestampName = mConfig.getMessageTimestampName();
+        	String messageTimestampSeparator = mConfig.getMessageTimestampNameSeparator();
+        	Object fieldValue = null;
+        	
+        	if (messageTimestampSeparator != null) {
+	        	String[] fields = messageTimestampName.split(messageTimestampSeparator);
```
jgagnon1: We use space for indentation (not tabs as in here)

mcehlert:
The reason for handling the separator was to also allow for non-nested keys that include . (dot).
{"foo.timestamp": "1405911096123", "bar.data":"test"}
Granted that's not a use case we have, but others might.

src/main/java/com/pinterest/secor/parser/DateMessageParser.java
```diff
@@ -52,7 +52,27 @@ public DateMessageParser(SecorConfig config) {
         String result[] = { defaultDate };
 
         if (jsonObject != null) {
-            Object fieldValue = jsonObject.get(mConfig.getMessageTimestampName());
+            String messageTimestampName = mConfig.getMessageTimestampName();
```
HenryCaiHaiying: This code is exactly the same as the one in JsonMessageParser, can they be consolidated in one place?  e.g. in MessageParser.java?

src/main/java/com/pinterest/secor/parser/DateMessageParser.java
```diff
@@ -52,7 +52,27 @@ public DateMessageParser(SecorConfig config) {
...
+            String messageTimestampSeparator = mConfig.getMessageTimestampNameSeparator();
+            Object fieldValue = null;
+
+            if (messageTimestampSeparator != null) {
+                String[] fields = messageTimestampName.split(messageTimestampSeparator);
+                Object finalValue = null;
+
+                for (int i=0; i < fields.length; i++) {
+                    if (!jsonObject.containsKey(fields[i])) { break; }
```
HenryCaiHaiying: You might want to log a warning in this case.

HenryCaiHaiying:
After you make the changes to the file, also make sure to use 'git squash' to squash them into one commit.

mcehlert:
Just added a fix: String.split() expects a regex, so the separator needed to be escaped via Pattern.quote().
Anything else I should do here?

src/main/java/com/pinterest/secor/parser/MessageParser.java
```diff
@@ -44,4 +50,30 @@ public ParsedMessage parse(Message message) throws Exception {
     }
 
     public abstract String[] extractPartitions(Message payload) throws Exception;
+    
+    public Object getJsonFieldValue(JSONObject jsonObject) {
+        Object fieldValue = null;
+        String messageTimestampName = mConfig.getMessageTimestampName();
+        String messageTimestampSeparator = mConfig.getMessageTimestampNameSeparator();
+        if (messageTimestampSeparator != null) {
+            String[] fields = messageTimestampName.split(Pattern.quote(messageTimestampSeparator));
```
HenryCaiHaiying: Since you need to create a pattern object, this is probably better done in the constructor/initializer of the MessageParser, rather than inside of loop of message parsing.

HenryCaiHaiying:
Looks fine, one small comment on when to initializing a few private variables.

mcehlert:
Splitting of the timestamp field has been moved to the constructor of MessageParser with the implication, that the properties "timestamp.name" and "timestamp.name.separator" need to be accessed even if MessageOffsetParser is used. Exception handling also added to the initializer.

HenryCaiHaiying:
The exception handling in the constructor is questionable, I think it's probably better to have it fail (throw up that RuntimeException) if the configuration is not right rather than try to continue.  People usually don't notice a warning message up in the beginning of the log.

Also github says: 'This branch has conflicts that must be resolved', can you take a look?

mcehlert:
If the exception is not handled, then
1) the test testCreateMessageParser in ReflectionUnitTest.java will fail because the empty mSecorConfig object does not contain the timestamp and separator fields which are then required during instantiation of every MessageParser subclass (here: OffsetMessageParser). This could be solved by adding empty fields to the test mSecorConfig object.
2) when using secor with MessageOffsetParser, an exception would be thrown if the config fields "timestamp.name" and "timestamp.name.separator" are not present (i.e. commented out) in the config. The point of MessageOffsetParser is to not read any timestamp fields inside the message, so it feels a bit off, that these options need to be present (even if not set to an actual value), I would think.
Any ideas for an elegant solution for this?

HenryCaiHaiying:
For 1) test can always be enhanced for the latest code behavior, adding empty fields in the test is fine with me;
For 2) The two new fields are present in the default properties file with empty value, so it should not get exception on that constructor.  If people comment out those two properties, they should be notified about the problem upfront and figure out a way to deal with it (either by adding those two properties or catch the exception from the constructor), I don't like the configuration issue/conflict lost in a big log file.

mcehlert:
Adjusted the test, removed exception handling and resolved merge conflict.

HenryCaiHaiying:
Looks good, thanks for the patience on going through several iterations.

PR #168, opened 2016-01-21
Change the url to access to qubole from http to https, looks like qub…

HenryCaiHaiying:
…ole now requires https access now.

Fix the error handling on response object to get better error report when the response is not a map.


pgarbacki:
r+

PR #170, opened 2016-01-26
Track daily and hourly stats separately

HenryCaiHaiying:
Currently both daily and hourly stats are published to the same TSD metric (topic=XXX, partition=XXX}, add the third dimension (consumer_group) to differentiate the two pipelines.


pgarbacki:
r+

PR #171, opened 2016-01-28
Fix a regresion on JsonMessageParser where the timestamp separator is…

HenryCaiHaiying:
… checked against null, not empty string

This was a recent checkin by open source community, looks like the check is not complete (only checked against null, not empty string), this caused the timestamp field cannot be retrieved.


pgarbacki:
r+

PR #172, opened 2016-02-01
Tighten the check on missing timestamp field for json message

HenryCaiHaiying:
We had a regresion last week when someone made a checkin which didn't extract the correct timestamp field which makes the json message passing through with timestamp information extracted correctly.

Tighten the check to make sure the timestamp field is always required for json messages.

Make this check dependdent on config property: message.timestamp.required, just in case some installations had broken messages but they don't want to change the behavior.


pgarbacki:
r+

PR #175, opened 2016-02-03
Added proxy support for S3UploadManager

vthakre:
This PR is for adding the proxy settings to enable accessing AWS S3 through proxy server 

@HenryCaiHaiying

### DIFF
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -237,6 +237,18 @@ public class SecorConfig {
     public String getAwsEndpoint() {
         return getString("aws.endpoint");
     }
+    
+    public boolean getAwsProxyEnabled(){
+    	return getBoolean("aws.proxy.isEnabled");
+    }
+    
+    public String getAwsProxyHttpHost() {
+        return getString("aws.proxy.http.host");
+    }
+    
+    public int getAwsProxyHttpPort() {
+        return getInt("aws.proxy.http.port");
+    }
 
     public String getAwsRegion() {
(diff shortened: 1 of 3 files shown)


src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -237,6 +237,18 @@ public String getAwsSecretKey() {
     public String getAwsEndpoint() {
         return getString("aws.endpoint");
     }
+    
+    public boolean getAwsProxyEnabled(){
+    	return getBoolean("aws.proxy.isEnabled", false);
```
HenryCaiHaiying: We usually want the config default specified in properties file, there is no need to provide a default in java code.

HenryCaiHaiying:
Looks fine, only a small comment on property default.

When you made the follow-up fix, can you squash two commits into one so it's cleaner in revision history.

src/main/java/com/pinterest/secor/common/SecorConfig.java
vthakre: for the boolean property, the method getBoolean expects 2 parameter, properyName and the defaultValue.

HenryCaiHaiying: Can you add a getBoolean method which doesn't require default value?  Or
directly calling mProperties.getBoolean()

vthakre: Added a method getBoolean with just propertyName. Also Squashed the commits

HenryCaiHaiying:
@vthakre Looks good, thanks for the patience on the iterations.

PR #178, opened 2016-02-12
Fix for custom compression codec

treff7es:
I tried to use lzo compression with secor but it always failed with null pointer exception because it did not find the set codec even if it was on the classpath.
I checked it and it is because it was not added to the CompressionCodecFactory's configuration. 
Due to this ccf.getCodecByClassName won't give back the set compression codec (except if it is gzip which is set by default) as it won't be in the hashmap (codecsByName) where it tries to find the set codec.

With the modification I did in the pull request lzo compression worked fine for us.


pgarbacki:
Thanks for fixing, @treff7es

PR #183, opened 2016-02-19
Add direct upload config flag for Google cloud storage, fixes #177.

### LINKED ISSUE #177: Google cloud storage insufficient data written
We have a secor setup where everything works with s3, but when switching to GS it throws

```
2016-02-11 23:45:10,502 [Thread-3] (com.pinterest.secor.common.FileRegistry:162) INFO  Deleting writer for path /data/secor/message_logs/partition/30946_12/organic-impressions/dt=2016-02-04/1_0_00000000000025099615
2016-02-11 23:45:10,503 [Thread-3] (com.pinterest.secor.uploader.GsUploadManager:73) INFO  uploading file /data/secor/message_logs/partition/30946_12/organic-impressions/dt=2016-02-04/1_0_00000000000025099615 to gs://trendmd-druid/raw/organic-impressions/dt=2016-02-04/1_0_00000000000025099615
Feb 11, 2016 11:45:11 PM com.twitter.common.zookeeper.DistributedLockImpl unlock
INFO: Cleaning up this locks ephemeral node.
Feb 11, 2016 11:45:11 PM com.twitter.common.zookeeper.DistributedLockImpl cleanup
INFO: Cleaning up!
2016-02-11 23:45:11,622 [Thread-3] (com.pinterest.secor.main.ConsumerMain$1:65) ERROR Thread Thread[Thread-3,5,main] failed
java.lang.RuntimeException: Failed to apply upload policy
    at com.pinterest.secor.consumer.Consumer.checkUploadPolicy(Consumer.java:112)
    at com.pinterest.secor.consumer.Consumer.run(Consumer.java:102)
Caused by: java.util.concurrent.ExecutionException: java.lang.RuntimeException: java.io.IOException: insufficient data written
    at java.util.concurrent.FutureTask.report(FutureTask.java:122)
    at java.util.concurrent.FutureTask.get(FutureTask.java:188)
    at com.pinterest.secor.uploader.FutureHandle.get(FutureHandle.java:34)
    at com.pinterest.secor.uploader.Uploader.uploadFiles(Uploader.java:95)
    at com.pinterest.secor.uploader.Uploader.checkTopicPartition(Uploader.java:186)
    at com.pinterest.secor.uploader.Uploader.applyPolicy(Uploader.java:207)
    at com.pinterest.secor.consumer.Consumer.checkUploadPolicy(Consumer.java:110)
    ... 1 more
Caused by: java.lang.RuntimeException: java.io.IOException: insufficient data written
    at com.pinterest.secor.uploader.GsUploadManager$1.run(GsUploadManager.java:94)
    at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471)
    at java.util.concurrent.FutureTask.run(FutureTask.java:262)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
    at java.lang.Thread.run(Thread.java:745)
Caused by: java.io.IOException: insufficient data written
    at sun.net.www.protocol.http.HttpURLConnection$StreamingOutputStream.close(HttpURLConnection.java:3228)
    at com.google.api.client.http.javanet.NetHttpRequest.execute(NetHttpRequest.java:81)
    at com.google.api.client.http.HttpRequest.execute(HttpRequest.java:972)
    at com.google.api.client.googleapis.media.MediaHttpUploader.executeCurrentRequestWithoutGZip(MediaHttpUploader.java:545)
    at com.google.api.client.googleapis.media.MediaHttpUploader.resumableUpload(MediaHttpUploader.java:417)
    at com.google.api.client.googleapis.media.MediaHttpUploader.upload(MediaHttpUploader.java:336)
    at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.executeUnparsed(AbstractGoogleClientRequest.java:427)
    at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.executeUnparsed(AbstractGoogleClientRequest.java:352)
    at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.execute(AbstractGoogleClientRequest.java:469)
    at com.pinterest.secor.uploader.GsUploadManager$1.run(GsUploadManager.java:92)
    ... 5 more
```

this only happens if the file size is larger than the chunk size (10MB), if we set `secor.max.file.size.bytes` to 7500000 it results in a file about 9.5MB and is uploaded to GS ok, but when we set `secor.max.file.size.bytes` to it 8000000 throws the exception (`secor.messages.per.second=10000`).

secor version: 0.16 (0.13 also)
jdk version: 7u91-2.6.3-1~deb8u1

eLod:


HenryCaiHaiying:
use getBoolean is probably better.  But the current getBoolean takes in the default value, we kind of want all the default value to come in from the properties file (not through java code).  How about you provide a getBoolean which doesn't take in default value and then you can use this new getBoolean

eLod:
@jgagnon1 @HenryCaiHaiying i specifically used the `getString.equals("true")` code because that is exactly how secor does it for `getSeperateContainersForTopics()` (`secor.swift.containers.for.each.topic`) and I wasn't sure if this is the recommended way. The other boolean setting i could find in this config is `getAwsProxyEnabled()` (which simply uses `getBoolean`), i used that first, but then found out the former (and again wasn't sure why these two different ways are both used).

@HenryCaiHaiying i'm a little perplexed by your comments:
- there is `getBoolean`, not just on `PropertiesConfiguration` but on `SecorConfig` itself as well,
- both has signatures for with and without default value,
- i don't understand why would we want to not have a default (this setting should not be mandatory, it should be off by default, e.g. opt-in)

That all said i'm fine with changing to one of the `getBoolean` forms, implementing anything else (especially adding new functionality) please feel free to do yourself.

HenryCaiHaiying:
For the properties default, secor want to keep a convention that all properties are mandatory (needs to be specified in properties file, and with the default value provided in properties file), we don't want to keep the default logic in two places (in properties file and in java code since they might diverge as times goes on).  Have every property explicitly listed in property file would force the adopter to take a look and examine the impact of any new property.  But sometimes this convention is not strictly enforced, that's when you see some inconsistencies.

There is one SecorConfig.getBoolean() which calls underlying PropertiesConfiguration.getBoolean(), it currently asks for the default.  I think we should also have a form of getBoolean which does not ask for default (i.e. the property needs to be present in property file).  This new getBoolean(propertyName) method would just call the corresponding getBoolean(propertyName) in PropertiesConfiguraration.  And you can use the new getBoolean() method in your code.

eLod:
@HenryCaiHaiying i don't understand. https://github.com/pinterest/secor/blob/master/src/main/java/com/pinterest/secor/common/SecorConfig.java#L407 vs https://github.com/pinterest/secor/blob/master/src/main/java/com/pinterest/secor/common/SecorConfig.java#L411. am i missing something here?

eLod:
@HenryCaiHaiying ok i'll use that method and amend my commit. not sure if you saw but i only added the config/setting to `src/main/config/secor.dev.gs.properties`, should i add it somewhere else? i checked `secor.common.properties` but i only saw mandatory GS settings there (e.g. `secor.gs.credentials.path` is only present in `secor.dev.gs.properties`, not really sure).

HenryCaiHaiying:
Didn't realize they used a separate properties file, we should just use one
properties file: secor.common.properties, that's the one used for
production environment.

eLod:
@HenryCaiHaiying ping (sry, but not sure if github is notifying you if i update commit)

jgagnon1:
LGTM

HenryCaiHaiying:
@eLod looks good to me, thanks for the contributions.
P.S.  The github actually didn't notify me your new update to the change, feel free to ping me if you don't get response.

PR #185, opened 2016-03-04
Add Iso8601MessageParser

jurriaan:
Enables proper Iso8601 parsing using javax.xml.bind.DatatypeConverter for Java 7 support.

This wasn't possible with the DateMessageParser because SimpleDateFormatter doesn't support optional elements.

### DIFF
--- a/src/main/java/com/pinterest/secor/parser/Iso8601MessageParser.java
+++ b/src/main/java/com/pinterest/secor/parser/Iso8601MessageParser.java
@@ -1,0 +1,69 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


src/main/java/com/pinterest/secor/parser/Iso8601MessageParser.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    public String[] extractPartitions(Message message) {
+        JSONObject jsonObject = (JSONObject) JSONValue.parse(message.getPayload());
+        String result[] = { defaultDate };
+
+        if (jsonObject != null) {
+            Object fieldValue = getJsonFieldValue(jsonObject);
+            if (fieldValue != null) {
```
HenryCaiHaiying: When fieldValue is not present, it will use the default, this might be error prone.  At least we need a warning logging here.

It might be better to introduce a config parameter to decide whether fallback to default or error out.

This also applies on when the field is available, but cannot be parsed.

jurriaan: Sure, I just copied DateMessageParser. Will add a warning here :)

HenryCaiHaiying:
@jurriaan Didn't realize you already made the change.  Github doesn't send notification when you make a new change/diff, feel free to ping me if you don't get response.

jurriaan:
Ah, great! Tnx :) will do

PR #193, opened 2016-03-21
making SecorConfig method public & use timezone from config for dateparser

lovenishgoyal:
Hi,

While I was trying to build project one of jUnit test case for Iso8601ParserTest.java class was failing. My timezone is GMT+05:30 and when it was processing one "timestamp":"2016-03-02T18:36:14+00:00" it is converting this into "2016-03-03T00:06:14+00:00"(As per GMT). I have modified the code based on configure timezone.

Please check and let me know in case of more clarification.

Thanks,
Lovenish

### DIFF
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -416,32 +416,32 @@ public class SecorConfig {
         return mProperties.getBoolean(name);
     }
 
-    private void checkProperty(String name) {
+    public void checkProperty(String name) {
         if (!mProperties.containsKey(name)) {
             throw new RuntimeException("Failed to find required configuration option '" +
                                        name + "'.");
         }
     }
 
-    private String getString(String name) {
+    public String getString(String name) {
         checkProperty(name);
         return mProperties.getString(name);
     }
 
-    private int getInt(String name) {
+    public int getInt(String name) {
         checkProperty(name);
         return mProperties.getInt(name);
     }
 
-    private int getInt(String name, int defaultValue) {
+    public int getInt(String name, int defaultValue) {
         return mProperties.getInt(name, defaultValue);
     }
 
-    private long getLong(String name) {
+    public long getLong(String name) {
         return mProperties.getLong(name);
     }
 
-    private String[] getStringArray(String name) {
+    public String[] getStringArray(String name) {
         return mProperties.getStringArray(name);
     }
 }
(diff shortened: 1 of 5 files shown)


lovenishgoyal:
Hi @HenryCaiHaiying , I am able to build the project with these new changes in my local, Could you please guide me about next step.
Thanks in advance

src/main/java/com/pinterest/secor/parser/DateMessageParser.java
```diff
@@ -41,9 +42,17 @@
     private static final Logger LOG = LoggerFactory.getLogger(DateMessageParser.class);
     protected static final String defaultDate = "dt=1970-01-01";
     protected static final String defaultFormatter = "yyyy-MM-dd";
-
+    SimpleDateFormat outputFormatter = new SimpleDateFormat(defaultFormatter);
```
HenryCaiHaiying: These variables should be either private or protected.

src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -416,32 +416,32 @@ public boolean getBoolean(String name) {
...
-    private void checkProperty(String name) {
+    public void checkProperty(String name) {
         if (!mProperties.containsKey(name)) {
             throw new RuntimeException("Failed to find required configuration option '" +
                                        name + "'.");
         }
     }
 
-    private String getString(String name) {
+    public String getString(String name) {
```
HenryCaiHaiying: Do you need to change these accessor methods to public?

HenryCaiHaiying:
Have you looked at the test failure in CI?

src/main/java/com/pinterest/secor/parser/DateMessageParser.java
lovenishgoyal: I'll make the change.

src/main/java/com/pinterest/secor/common/SecorConfig.java
lovenishgoyal: these are common methods so we can use from our customized classes

lovenishgoyal:
CI is not showing any error, Build got failed while it was downloading jars. Test cases did not even start.
https://travis-ci.org/pinterest/secor/jobs/117530772

Can we start the build once again ?

lovenishgoyal:
@HenryCaiHaiying - Ping

HenryCaiHaiying:
Restarting the CI build

lovenishgoyal:
Hi @HenryCaiHaiying, I have changed Access Modifier for variables.

HenryCaiHaiying:
Looks good, thanks.

PR #195, opened 2016-03-27
Avro Message Parser

ryananguiano:
I think that this should fix the Avro parser but I am having a little trouble setting up the tests. Any suggestions how to mock the schema registry?


HenryCaiHaiying:
Kick off the CI test again.

Igosuki:
The problem with this PR is that it does not differentiate between normal Avro and Confluent Avro messages (with a byte + 4 bytes for a 32 bit schema id). I think there should be two AvroUtil classes, one for plain Avro and one for Schema-Registry based confluent messages.

HenryCaiHaiying:

PR #197, opened 2016-04-09
add Aws Role support

yyejun:
Added a new config aws.role for cross account STSAssumeRole config.


HenryCaiHaiying:
looks good, thanks.

PR #198, opened 2016-04-18
Reorder handling of codec and file-extension

mcehlert:
In the current state, if the config defines both compression codec and file extension, the compression codec is not applied. 
Setting a compression codec should not rely on file extension to NOT be set. Especially since the comments in the config file state "gz" as an example for `secor.file.extension`


HenryCaiHaiying:
Thanks for the fix.

PR #202, opened 2016-04-19
upload files to a new s3 path after a given migration date

shawnsnguyen:
This change will allow the uploader to upload files to a new S3 path after a given date (labeled as the migration date) in the secor configs.

### DIFF
--- a/src/main/config/secor.common.properties
+++ b/src/main/config/secor.common.properties
@@ -291,3 +291,13 @@ secor.parser.timezone=UTC
 
 # Transformer class that transform message accordingly.  
 secor.message.transformer.class=com.pinterest.secor.transformer.IdentityMessageTransformer
+
+# Set below property to true if you want to have the md5hash appended to your s3 path.
+# This helps for better partitioning of the data on s3. Which gives better performance while reading and writing on s3
+secor.s3.prefix.md5hash=false
+
+# After the given date, secor will upload files to the supplied s3 alternative path
+secor.s3.alter.path.date=
+
+# An alternative S3 path for secor to upload files to
+secor.s3.alternative.path=
--- a/src/main/config/secor.common.properties
+++ b/src/main/config/secor.common.properties
@@ -301,4 +301,3 @@ secor.s3.alter.path.date=
 
 # An alternative S3 path for secor to upload files to
 secor.s3.alternative.path=
-
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -171,6 +171,14 @@ public class SecorConfig {
         return getString("secor.s3.path");
     }
 
+    public String getS3MigrationPath() {
+        return getString("secor.s3.migration.path");
+    }
+
+    public String getS3MigrationDate() {
+        return getString("secor.s3.migration.date");
+    }
+
     public String getS3Prefix() {
         return getS3FileSystem() + "://" + getS3Bucket() + "/" + getS3Path();
     }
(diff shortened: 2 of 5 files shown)


src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -171,6 +171,14 @@ public String getS3Path() {
         return getString("secor.s3.path");
     }
 
+    public String getS3MigrationPath() {
```
HenryCaiHaiying: New properties needs to have a line in src/main/config/secor.common.properties.

Maybe the property name can be called S3AlternatePath and S3AlthertePathDate

src/main/java/com/pinterest/secor/uploader/S3UploadManager.java
```diff
@@ -117,11 +119,23 @@ public void refresh() {}
 
     public Handle<?> upload(LogFilePath localPath) throws Exception {
         String s3Bucket = mConfig.getS3Bucket();
-        String s3Key = localPath.withPrefix(mConfig.getS3Path()).getLogFilePath();
+        Date s3MigrationDate = null;
+        String s3Path;
+        try {
+            s3MigrationDate = new SimpleDateFormat("yyyy-MM-dd").parse(mConfig.getS3MigrationDate());
```
HenryCaiHaiying: S3MigrateDate and S3MigratePath should be evaluated in UploadManager's construction time.

src/main/java/com/pinterest/secor/uploader/S3UploadManager.java
```diff
@@ -117,11 +119,23 @@ public void refresh() {}
...
+        } catch (RuntimeException e) {
+            // optional migration date, so do nothing if it's not defined
```
HenryCaiHaiying: I think once you had an entry in secor.common.properties, you probably won't see that RuntimeException.

src/main/java/com/pinterest/secor/uploader/S3UploadManager.java
```diff
@@ -117,11 +119,23 @@ public void refresh() {}
...
+        }
+        if (s3MigrationDate != null && !s3MigrationDate.after(new Date())) {
+            s3Path = mConfig.getS3MigrationPath();
```
HenryCaiHaiying: Probably leave a INFO level logging when the s3 path switched.

src/main/java/com/pinterest/secor/uploader/S3UploadManager.java
HenryCaiHaiying: Also there is HadoopS3UploadManager, maybe you also need to change there.

src/main/config/secor.common.properties
```diff
@@ -294,4 +294,10 @@ secor.message.transformer.class=com.pinterest.secor.transformer.IdentityMessageT
 
 # Set below property to true if you want to have the md5hash appended to your s3 path.
 # This helps for better partitioning of the data on s3. Which gives better performance while reading and writing on s3
-secor.s3.prefix.md5hash=false
\ No newline at end of file
+secor.s3.prefix.md5hash=false
+
+# After the given date, secor will upload files to the supplied s3 alternative path
+secor.s3.alter_path.date=
```
HenryCaiHaiying: We don't use '_' in the parameter name, you can either call it 'alter.path.date' or 'alternative.path.date'

src/main/java/com/pinterest/secor/uploader/S3UploadManager.java
```diff
@@ -69,7 +71,11 @@
     private static final String S3 = "S3";
     private static final String CUSTOMER = "customer";
 
+    private final String S3_PATH;
```
HenryCaiHaiying: This instance variable should not be uppercased.  The upper case are for constants.

PR #204, opened 2016-04-26
kafka topic blacklist

yyejun:
Support kafka consumer blacklist.

### DIFF
--- a/src/main/java/com/pinterest/secor/reader/MessageReader.java
+++ b/src/main/java/com/pinterest/secor/reader/MessageReader.java
@@ -29,6 +29,7 @@ import kafka.consumer.ConsumerIterator;
 import kafka.consumer.KafkaStream;
 import kafka.consumer.TopicFilter;
 import kafka.consumer.Whitelist;
+import kafka.consumer.Blacklist;
 import kafka.javaapi.consumer.ConsumerConnector;
 import kafka.message.MessageAndMetadata;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


src/main/java/com/pinterest/secor/reader/MessageReader.java
```diff
@@ -65,7 +66,12 @@ public MessageReader(SecorConfig config, OffsetTracker offsetTracker) throws
 
         mConsumerConnector = Consumer.createJavaConsumerConnector(createConsumerConfig());
 
-        TopicFilter topicFilter = new Whitelist(mConfig.getKafkaTopicFilter());
+        if (!mConfig.getKafkaTopicBlacklist().isEmpty() && !mConfig.getKafkaTopicFilter().isEmpty()) {
+            throw new RuntimeException("Topic filter and blacklist cannot be both specified.");
+        }
+        TopicFilter topicFilter = !mConfig.getKafkaTopicBlacklist().isEmpty()? new Blacklist(mConfig.getKafkaTopicBlacklist()):
```
HenryCaiHaiying: Do you need to check both null and empty string?

yyejun: If both empty, it will behavior same as current code return a WhileList(""). Not sure if it matches everything or nothing.

HenryCaiHaiying: Actually I don't know whether you will get null from mConfig.getKafkaTopicBlacklist(), you probably will get NPE is the property is not specified.  Just do a test on not specifying that new property, if that looks good, I can merge in your PR.

yyejun: If it is not specified instead of empty string, it will not pass https://github.com/pinterest/secor/blob/master/src/main/java/com/pinterest/secor/common/SecorConfig.java#L458.
I also tested both empty value, it runs but empty WhiteList.

HenryCaiHaiying:
Sounds good, thanks for the enhancement.

PR #207, opened 2016-04-28
Make Uploader pluggable through secor.properties

lucamilanesio:
Allow custom logic to be applied to flush locally stored topic files
to the underlying storage system.

Currently, only a single logic can be applied, and it is coded in the Uploader class.
By making the class pluggable and configurable through secor.upload.class property
we can just override the applyPolicy() and customize the logic on how to flush data.

This change is fully backward compatible. When property is not configured the Uploader
class is instantiated as before.

### DIFF
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -215,6 +215,10 @@ public class SecorConfig {
         return getString("secor.message.parser.class");
     }
 
+    public String getUploaderClass() {
+        return getString("secor.upload.class", "com.pinterest.secor.uploader.Uploader");
+    }
+
     public String getUploadManagerClass() {
         return getString("secor.upload.manager.class");
     }
@@ -461,6 +465,10 @@ public class SecorConfig {
     public String getString(String name) {
         checkProperty(name);
         return mProperties.getString(name);
+    }
+
+    public String getString(String name, String defaultValue) {
+        return mProperties.getString(name, defaultValue);
     }
 
     public int getInt(String name) {
--- a/src/main/java/com/pinterest/secor/consumer/Consumer.java
+++ b/src/main/java/com/pinterest/secor/consumer/Consumer.java
@@ -71,7 +71,8 @@ public class Consumer extends Thread {
         FileRegistry fileRegistry = new FileRegistry(mConfig);
         UploadManager uploadManager = ReflectionUtil.createUploadManager(mConfig.getUploadManagerClass(), mConfig);
 
-        mUploader = new Uploader(mConfig, mOffsetTracker, fileRegistry, uploadManager);
+        mUploader = ReflectionUtil.createUploader(mConfig.getUploaderClass());
+        mUploader.init(mConfig, mOffsetTracker, fileRegistry, uploadManager);
         mMessageWriter = new MessageWriter(mConfig, mOffsetTracker, fileRegistry);
         mMessageParser = ReflectionUtil.createMessageParser(mConfig.getMessageParserClass(), mConfig);
         mMessageTransformer =  ReflectionUtil.createMessageTransformer(mConfig.getMessageTransformerClass(), mConfig);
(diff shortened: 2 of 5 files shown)


src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -17,6 +17,7 @@
 package com.pinterest.secor.common;
 
 import com.google.api.client.repackaged.com.google.common.base.Strings;
+import com.pinterest.secor.uploader.Uploader;
```
HenryCaiHaiying: Doesn't quite like this added dependency to secor.uploader package, I think you can have this property have a default value in secor.common.properties (there are quite a few config property specifying classname in secor.common.properties already)

src/main/java/com/pinterest/secor/consumer/Consumer.java
```diff
@@ -71,7 +71,7 @@ private void init() throws Exception {
         FileRegistry fileRegistry = new FileRegistry(mConfig);
         UploadManager uploadManager = ReflectionUtil.createUploadManager(mConfig.getUploadManagerClass(), mConfig);
 
-        mUploader = new Uploader(mConfig, mOffsetTracker, fileRegistry, uploadManager);
+        mUploader = ReflectionUtil.createUploader(mConfig.getUploaderClass(), mConfig, mOffsetTracker, fileRegistry, uploadManager);
```
HenryCaiHaiying: If we use reflection to create instance, the standard way is to create the object using no-arg constructor through reflection, and call a configure method with the rest of the arguments.

src/main/java/com/pinterest/secor/common/SecorConfig.java
lucamilanesio: @HenryCaiHaiying thanks for the feedback, makes sense, will include in my next amended commit.

src/main/java/com/pinterest/secor/consumer/Consumer.java
lucamilanesio: @HenryCaiHaiying I was about to take that approach, however, other methods (e.g. createMessageTransformer) invoke the constructor with parameters. I would change it to newInstance + init anyway as you requested.

lucamilanesio:
@HenryCaiHaiying all changes you requested have been applied, could you have a second look please? Thank you in advance. Luca.

src/main/java/com/pinterest/secor/uploader/Uploader.java
```diff
@@ -47,16 +47,17 @@
     private ZookeeperConnector mZookeeperConnector;
     private UploadManager mUploadManager;
 
-    public Uploader(SecorConfig config, OffsetTracker offsetTracker, FileRegistry fileRegistry,
-                    UploadManager uploadManager) {
-        this(config, offsetTracker, fileRegistry, uploadManager,
-             new ZookeeperConnector(config));
+
```
HenryCaiHaiying: Can you add the javadoc to this 'init' method?

src/main/java/com/pinterest/secor/util/ReflectionUtil.java
```diff
@@ -61,6 +64,32 @@ public static UploadManager createUploadManager(String className,
     }
 
     /**
+     * Create an Uploader from its fully qualified class name.
+     *
+     * The class passed in by name must be assignable to Uploader
```
HenryCaiHaiying: Please update the javadoc since you don't use these 4 arguments now.

src/main/java/com/pinterest/secor/uploader/Uploader.java
lucamilanesio: Will do, even though there was no JavaDoc on the previous constructor ... but I can deduct what is what from the code :-)

src/main/java/com/pinterest/secor/util/ReflectionUtil.java
lucamilanesio: Done.

lucamilanesio:
@HenryCaiHaiying thanks for the extra reviews, amended the commit and rebased on current master.

lucamilanesio:
Build is green, hope we should be good to go 😄

HenryCaiHaiying:

Issue #78, opened 2015-02-23
Locale-dependent test


andrioni:
Hello! [This DateMessageParser test](https://github.com/pinterest/secor/blob/master/src/test/java/com/pinterest/secor/parser/DateMessageParserTest.java#L53-L55) is locale-dependent, and fails in non-English locales (which is how I discovered it!).

What would the best solution for this issue? Adding a new locale parameter (e.g. `message.timestamp.locale`) or just removing the test?

pgarbacki:
Although it is a bit ugly from the stylistic point of view, you could
change the default locale
<http://docs.oracle.com/javase/7/docs/api/java/util/Locale.html#setDefault(java.util.Locale)>
for that test, and then change it back.

pgarbacki:
@zago would you mind taking a look at this test?

PR #209, opened 2016-05-03
fix Locale in DateMessageParser, fix Locale dependency

nguyenvanthan:
Fix Locale in DateMessageParser.
The Locale should be set in a config file or unit test should not be local dependent.

Should fix https://github.com/pinterest/secor/issues/78

### DIFF
--- a/src/main/java/com/pinterest/secor/parser/DateMessageParser.java
+++ b/src/main/java/com/pinterest/secor/parser/DateMessageParser.java
@@ -18,6 +18,7 @@ package com.pinterest.secor.parser;
 
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.Locale;
 import java.util.TimeZone;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


HenryCaiHaiying:
Will this break the program if people are running secor in Europe?

nguyenvanthan:
It depends if you use number representation or letter representation in your date pattern.
Example: ```dd-MM-yyyy``` will be fine but if you use  ```dd MMMMM yyyy``` it will expect the month as a word based on a Locale.
Maybe you should ask users to use only digits in date pattern.

HenryCaiHaiying:
How about you make this defaultLocal (and also defaultFormatter) a configuration parameter coming through secor.common.properties?

nguyenvanthan:
Maybe i can add ```message.timestamp.input.pattern.locale.language```, default to "en" and where language is "an ISO 639 alpha-2 or alpha-3 language code, or a language subtag up to 8 characters in length" ?

HenryCaiHaiying:
This seems quite complicated.  I think as long as you allow people specify their own 'defaultLocale'/defaultFormatter, that should be fine.

HenryCaiHaiying:

PR #211, opened 2016-05-04
Add support for AWS S3 Accelerate as an option for S3 Uploader

ranbirchawla:
I've add a new property designed to enable the new S3 Accelerate feature. To do so I've changed the SDK versions to 1.10.74 and added logic.

Right now since S3 Accelerate will over-write the endpoint used I organized the logic to default to false for S3 Accelerate if an endpoint is specified. If that is left blank and the region is specified and the property is true the upload hander will use the S3 accelerate feature. If it is left false the region will still be set on the client.

The feature must also be enabled on the bucket as well to function properly.


HenryCaiHaiying:
The change looks fine, how fast is S3 accelerate?

Can you squash the 4 commits into one?

HenryCaiHaiying:

PR #212, opened 2016-05-06
Change prefix path to only include the alternative path, not the entire bucket in the alternate path

shawnsnguyen:


shawnsnguyen:
Will rebase, and submit a new PR

PR #215, opened 2016-05-10
Expose existing Log file writers

lucamilanesio:
Make the existing writers of the log files accessible from the outside
as they may be needed for custom uploaders.


HenryCaiHaiying:
Thanks for the contribution

PR #216, opened 2016-05-10
Update partition finalizer to use full path prefix instead of just the prefix

shawnsnguyen:
This change adjusts the path prefix in the partition finalizer to include the uri and bucket, rather than the prefix alone.


HenryCaiHaiying:
Please add description for the PR

PR #217, opened 2016-05-13
Update feature in READme.md

lovenishgoyal:
Added some feature like msg level transformation, partition level details


HenryCaiHaiying:
Thanks for the contribution.

PR #220, opened 2016-06-10
Remove redundant exec-maven-plugin

justinmills:
It causes a warning in the build and is generating the properties file
in the wrong place.


HenryCaiHaiying:

PR #221, opened 2016-06-13
Allow for git or env var to drive build.properties generation

justinmills:
This allows you to set an env var (SOURCE_VERSION) instead of relying on
git command line to be present to generate the build.properties file.
Very useful if you deploy to an environment where git may not be
present, or the checkout may have it's ties back to the git repo from
whence it came severed, as is the case when deploying to Heroku.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -331,28 +331,6 @@
                 </executions>
             </plugin>
             <plugin>
-                <groupId>org.codehaus.mojo</groupId>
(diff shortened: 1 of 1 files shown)


HenryCaiHaiying:
I will hold the merge of this PR since secor at head seems having some issues on log4j.

HenryCaiHaiying:

Issue #224, opened 2016-06-28
Google cloud storage credential file config should be optional


yunjing:
`secor.gs.credentials.path` should be optional because if secor is running on Google Cloud VMs, no credential is needed. In fact, the code here https://github.com/pinterest/secor/blob/master/src/main/java/com/pinterest/secor/uploader/GsUploadManager.java#L114 already checks if the file config is null.

HenryCaiHaiying:
Can you prepare a PR for this?

Thanks.

yunjing:
https://github.com/pinterest/secor/pull/225

PR #225, opened 2016-06-28
Make secor.gs.credentials.path optional

yunjing:
This is discussed in https://github.com/pinterest/secor/issues/224

### DIFF
--- a/src/main/config/secor.common.properties
+++ b/src/main/config/secor.common.properties
@@ -113,6 +113,11 @@ secor.gs.path=data
 # https://cloud.google.com/storage/docs/json_api/v1/how-tos/upload
 secor.gs.upload.direct=false
 
+# Application credentials configuration file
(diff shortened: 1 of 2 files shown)


src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -397,7 +397,7 @@ public String getZookeeperPath() {
     }
 
     public String getGsCredentialsPath() {
-        return getString("secor.gs.credentials.path");
```
HenryCaiHaiying: We usually prefer people providing default value through properties file, we want to make the default value more explicit.  Can you provide the default value in secor.common.properties?

yunjing:
Done.

HenryCaiHaiying:
Looks good, thanks.

PR #226, opened 2016-06-29
Added support for writing protobuf messages directly to Parquet

spektom:


HenryCaiHaiying:
Some high level questions:

- Can ProtoBuff and Parquet be decoupled so other path (e.g. Thrift) can also use Parquet, can we configure people to choose between SequenceFile and ParquetFile output format?
- What's the schema look like in Parquet file, how does that corresponds to the schema on message side?
- Pass in more arguments through reflection is convoluated, is there a better way of doing that?

spektom:
1. I thought about decoupling, but there's no much to decouple. If you look at the `ProtobufParquetFileWriter` and `ProtobufParquetFileReader` implementations, they are very difficult to make abstract, since quite every line is implementation (Protobuf) specific. It's easy to create Thrift or Avro implementation using relevant `ParquetWriter` or `ParquetReader` implementations. Anyway, I think premature decoupling is not worth in this case.

2. Protobuf schema is propagated to the Parquet format using `ProtoWriteSupport` class, where protobuf types are mapped to the relevant Parquet column types. Message schema is determined from the message class, which contains the schema in its static descriptor data fields.

3. I didn't find a better way than calling static method `parseFrom` of protobuf class for creating message instance out of byte array (if this is what you mean). It's a bit of an overhead, I agree, but it seems like there's no way to avoid reflection.

I'll be glad to hear your suggestions, if you have any.

spektom:
Guys, do you need help in maintaining this repository? I've noticed tens of open pull requests, which were left without attention from your side, so I wonder whether you are very busy or it's just these PR are out of scope. In the first case I'm offering you my help, in the latter one I'll just continue maintaing my own fork, and publish maven artifacts out of it.

HenryCaiHaiying:
My main concern is the on the reflectionUtil on creating ReaderWriterFactory with optional SecorConfig, but I couldn't figure out a better way of not breaking API compatibility.

The other concern I have is you can only specify one MessageClass for the ReaderWriterFactory, it seems we need a map between a topic and the message class, (the topic name is retrievable from LogFilePath), maybe the config in SecorConfig really should be a map?

HenryCaiHaiying:
Looks good, thanks for the patience to get this through, it's a great addition to secor.

nguyenvanthan:
Hello, it's out of scope maybe but in the case that you have any clue. 
@spektom do you know how parquet or Secor transform protobuf enum values ?
I have a protobuf enum like this : 
``` 
enum TransactionStatus {
        ACCEPTED = 0;
        REJECTED = 1;
}
```
I use Secor to transform to parquet file and in the parquet file, i have a value like this : 
```QUNDRVBURUQ=```
I try to convert the value to something that look likes my enum value ....

spektom:
Hello Christian,

Enums are encoded using Binary Enum types in Parquet:
https://github.com/Parquet/parquet-mr/blob/master/parquet-protobuf/src/main/java/parquet/proto/ProtoSchemaConverter.java#L101

Spark sees them as String though, because there's no such thing as Binary
Enum type in Spark.

It's weird, but in your case values are written in a base64 encoded format
(decoding "QUNDRVBURUQ=" would give "ACCEPTED").
How do you retrieve values from Parquet?

Regards,
Michael

nguyenvanthan:
Thanks @spektom, i don't check in base64.
I read directly from parquet files with parquet tools binary.
Now i have to find where it transform to base64, i would like to have the protobuf enum int value or int index...

PR #231, opened 2016-07-19
Adding argument reading to LogVerifier

prasincs:
It was failing with NullPointerException because it wasn't looking for an argument here. I can add a test too if required but this seemed obvious enough.

### DIFF
--- a/src/main/java/com/pinterest/secor/main/LogFileVerifierMain.java
+++ b/src/main/java/com/pinterest/secor/main/LogFileVerifierMain.java
@@ -49,11 +49,13 @@ public class LogFileVerifierMain {
                .create("t"));
         options.addOption(OptionBuilder.withLongOpt("start_offset")
                .withDescription("offset identifying the first set of files to check")
+               .hasArg()
                .withArgName("<offset>")
                .withType(Long.class)
                .create("s"));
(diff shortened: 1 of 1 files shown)


src/main/java/com/pinterest/secor/main/LogFileVerifierMain.java
```diff
@@ -49,11 +49,13 @@ private static CommandLine parseArgs(String[] args) throws ParseException {
                .create("t"));
         options.addOption(OptionBuilder.withLongOpt("start_offset")
                .withDescription("offset identifying the first set of files to check")
+               .hasArg()
```
HenryCaiHaiying: Does this make this arg required argument?  What exception stack trace did you get on the old code?

prasincs: it was just returning null and failing to parse before. So basically NullPointerException. 
It doesn't seem to be making it a required argument because I have jobs that are running without the start and end arguments.

HenryCaiHaiying: I think you are right, hasArg just to indicate it has an argument, not to make it a required argument.

HenryCaiHaiying:
Thanks for the contribution.

PR #236, opened 2016-07-29
Uploads at minute mark for selected topics

prasincs:
If secor is started at something other than the top of the hour mark or some other desired time, I've found that it doesn't upload at the desired times. We have operational requirements on some topics to have data at or close to the top of the hour mark. So this feature enables to start/restart secor whenever and be able to upload the files upto that point.

- This is restricted to particular topic filters that may require files
to be available at particular time for batch jobs
- Adds two configs:
  - secor.kafka.upload_at_minute_mark.topic_filter
  - secor.upload.minute_mark

### DIFF
--- a/src/main/java/com/pinterest/secor/uploader/Uploader.java
+++ b/src/main/java/com/pinterest/secor/uploader/Uploader.java
@@ -27,6 +27,7 @@ import com.pinterest.secor.util.ReflectionUtil;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.hadoop.io.compress.CompressionCodec;
+import org.joda.time.DateTime;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/pinterest/secor/uploader/Uploader.java
```diff
@@ -181,12 +182,31 @@ protected void trimFiles(TopicPartition topicPartition, long startOffset) throws
...
+        return false;
+    }
+
     private void checkTopicPartition(TopicPartition topicPartition) throws Exception {
         final long size = mFileRegistry.getSize(topicPartition);
         final long modificationAgeSec = mFileRegistry.getModificationAgeSec(topicPartition);
         LOG.debug("size: " + size + " modificationAge: " + modificationAgeSec);
         if (size >= mConfig.getMaxFileSizeBytes() ||
-                modificationAgeSec >= mConfig.getMaxFileAgeSeconds()) {
+                modificationAgeSec >= mConfig.getMaxFileAgeSeconds() ||
```
HenryCaiHaiying: Is this because this is a slow volume topic?  Otherwise modificationAgeSec would trigger the upload too.

prasincs: Yeah, some topics/environments are pretty slow volume and may not have all the data by the time their respective jobs run. I considered an alternate solution to just add a REST endpoint to flush to S3 whenever but that required changing more components than this solution.

src/main/java/com/pinterest/secor/uploader/Uploader.java
```diff
@@ -181,12 +182,31 @@ protected void trimFiles(TopicPartition topicPartition, long startOffset) throws
...
 
+    /***
+     * If the topic is in the list of topics to upload at a specific time. For example at a minute mark.
+     * @param topicPartition
+     * @return
+     * @throws Exception
+     */
+    private boolean isRequiredToUploadAtTime(TopicPartition topicPartition) throws Exception{
+        final String topic = topicPartition.getTopic();
+        final String topicFilter = mConfig.getKafkaTopicUploadAtMinuteMarkFilter();
```
HenryCaiHaiying: topicFilter initialization can be done in the constructor of the class.

src/main/java/com/pinterest/secor/uploader/Uploader.java
```diff
@@ -181,12 +182,31 @@ protected void trimFiles(TopicPartition topicPartition, long startOffset) throws
...
+        if (topicFilter == null || topicFilter.isEmpty()){ return false; }
```
HenryCaiHaiying: 'return false' should take it's own line for code format.

HenryCaiHaiying:

Issue #243, opened 2016-08-03
Iso8601MessageParser does not create a correct logger instance for its class


agolotin:
This is very minor, but if you look at [line 39 Iso8601MessageParser.java](https://github.com/pinterest/secor/blob/52a2a8b0e1bf6019177e24dc53d547464543043c/src/main/java/com/pinterest/secor/parser/Iso8601MessageParser.java#L39) it should be ```LoggerFactory.getLogger(Iso8601MessageParser.class)``` instead as far as I understand it...

HenryCaiHaiying:
Good catch, would you like to submit a PR for the fix?

KevinGrandon:
@HenryCaiHaiying Can this be closed now?

HenryCaiHaiying:
This is fixed already.

PR #244, opened 2016-08-03
Get the Iso8601 message parser to create correct log instance

agolotin:
In reference to #243


HenryCaiHaiying:
Thanks for the contribution.

Issue #232, opened 2016-07-21
Failed to apply upload policy when s3a is used.


oche191:
Hello. I've already found couple of similar issues like that e.g https://github.com/pinterest/secor/issues/160

When i am trying to use s3a to write to some bucket located in eu-central or eu-west i am always getting following error:

> `java.lang.RuntimeException: Failed to apply upload policy
>   at com.pinterest.secor.consumer.Consumer.checkUploadPolicy(Consumer.java:116)
>   at com.pinterest.secor.consumer.Consumer.run(Consumer.java:106)
> 
> Caused by: java.util.concurrent.ExecutionException: java.lang.NoSuchMethodError: com.amazonaws.services.s3.transfer.TransferManager.<init>(Lcom/amazonaws/services/s3/AmazonS3;Ljava/util/concurrent/ThreadPoolExecutor;)V
>   at java.util.concurrent.FutureTask.report(FutureTask.java:122)
>   at java.util.concurrent.FutureTask.get(FutureTask.java:192)
>   at com.pinterest.secor.uploader.FutureHandle.get(FutureHandle.java:34)
>   at com.pinterest.secor.uploader.Uploader.uploadFiles(Uploader.java:104)
>   at com.pinterest.secor.uploader.Uploader.checkTopicPartition(Uploader.java:196)
>   at com.pinterest.secor.uploader.Uploader.applyPolicy(Uploader.java:228)
>   at com.pinterest.secor.consumer.Consumer.checkUploadPolicy(Consumer.java:114)
>   ... 1 more
> 
> Caused by: java.lang.NoSuchMethodError: com.amazonaws.services.s3.transfer.TransferManager.<init>(Lcom/amazonaws/services/s3/AmazonS3;Ljava/util/concurrent/ThreadPoolExecutor;)V
>   at org.apache.hadoop.fs.s3a.S3AFileSystem.initialize(S3AFileSystem.java:287)
>   at org.apache.hadoop.fs.FileSystem.createFileSystem(FileSystem.java:2653)
>   at org.apache.hadoop.fs.FileSystem.access$200(FileSystem.java:92)
>   at org.apache.hadoop.fs.FileSystem$Cache.getInternal(FileSystem.java:2687)
>   at org.apache.hadoop.fs.FileSystem$Cache.get(FileSystem.java:2669)
>   at org.apache.hadoop.fs.FileSystem.get(FileSystem.java:371)
[...]

PR #253, opened 2016-09-22
Resolve hadoop and aws conflict from #232

joshua-rutherford:
Not sure on the policies regarding the use of alpha/snapshot dependencies in snapshots here, but the Hadoop s3a file system requires that that the AWS artifact versions match exactly as stated [here](https://wiki.apache.org/hadoop/AmazonS3):

> S3A requires the exact version of the amazon-aws-sdk against which Hadoop was built 
> (and is bundled with). If you try to upgrade the library by dropping in a later version, things 
> will break.

Thus, this pull request moves up to Hadoop 3.0.0-alpha and back to AWS 1.10.6.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,6 +51,8 @@
         <maven.compiler.target>1.6</maven.compiler.target>
         <encoding>UTF-8</encoding>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <version.aws>1.10.6</version.aws>
+        <version.hadoop>3.0.0-alpha1</version.hadoop>
     </properties>
 
     <distributionManagement>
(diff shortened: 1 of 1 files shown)


joshua-rutherford:
Actually, this is probably a major version bump (if you semantically version) since you'll no longer be able to support Java 7.

HenryCaiHaiying:
This change is probably too big for us to take now, we still need to support java 1.7 and taking hadoop alpha version is not a good practice for us.

joshua-rutherford:
Alright, I've looked at this and determined there is no backwards compatible solution to get support for s3a while depending upon hadoop for file system access.  Either you upgrade to hadoop 3+ and lose Java 7 or you downgrade to aws-java-sdk 1.7.4 and lose server side encryption.  I'll have to run from my own fork until you are ready to make the leap.

HenryCaiHaiying:
We will revisit this when hadoop 3 is more stable and people are ready to
cut java 7, thanks for the investigation.

PR #257, opened 2016-09-30
made auto.offset.reset configurable

avivian:
I have a use case where I need to start consuming from the largest offset rather than the smallest.


HenryCaiHaiying:
OK with the approach.  The convention of the new property is:
1. add it to secor.common.properties
2. for the default value, instead of fallback in java code, we want to read it from properties file which means just use getString"kafka.auto.offset.reset") without default value and have user specify the default value in property file.

HenryCaiHaiying:
I am closing this one since it's a dup

PR #259, opened 2016-10-03
added timeout to qubole calls

shawnsnguyen:
Since calls to qubole can hang, we want to add a timeout option to the add partition call so that it doesnt block the partition finalizer. Also change method to return a code depending on success of call so that upstream calls can retry if needed.

### DIFF
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -376,6 +376,8 @@ public class SecorConfig {
         return mProperties.getString(key, null);
     }
 
+    public boolean getSkipQuboleAddPartition() { return getBoolean("secor.skip.qubole.add.partition", false); }
+
     public String getCompressionCodec() {
         return getString("secor.compression.codec");
     }
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -506,10 +506,6 @@ public class SecorConfig {
         return mProperties.getLong(name);
     }
 
-    public long getLong(String name, long defaultValue) {
-        return mProperties.getLong(name, defaultValue);
-    }
-
     public String[] getStringArray(String name) {
         return mProperties.getStringArray(name);
     }
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -383,11 +383,11 @@ public class SecorConfig {
     }
 
     public boolean getQuboleEnabled() {
-        return getBoolean("secor.enable.qubole", true);
+        return getBoolean("secor.enable.qubole");
     }
 
     public long getQuboleTimeoutMs() {
-        return getLong("secor.qubole.timeout.ms", 300000);
+        return getLong("secor.qubole.timeout.ms");
     }
 
     public String getCompressionCodec() {
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -376,7 +376,13 @@ public class SecorConfig {
         return mProperties.getString(key, null);
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/pinterest/secor/parser/QuboleClient.java
```diff
@@ -94,24 +96,28 @@ private int query(String query) throws IOException {
...
     }
 
-    public void addPartition(String table, String partition) throws IOException,
+    public int addPartition(String table, String partition) throws IOException,
             InterruptedException {
         String queryStr = "ALTER TABLE " + table + " ADD IF NOT EXISTS PARTITION (" + partition +
             ")";
         int commandId = query(queryStr);
-        waitForCompletion(commandId);
+        return waitForCompletion(commandId, MAX_QUBOLE_WAIT_TIME_MS);
```
HenryCaiHaiying: How do you use the return code?

src/main/java/com/pinterest/secor/parser/QuboleClient.java
```diff
@@ -94,24 +96,28 @@ private int query(String query) throws IOException {
...
             if (response.get("status").equals("done")) {
-                return;
+                return 0;
             }
             System.out.println("waiting 3 seconds for results of query " + commandId +
                                ". Current status " + response.get("status"));
             Thread.sleep(3000);
         }
+
+        return 1; // qubole call failed to return within timeout
```
HenryCaiHaiying: Should it be -1 return on error or throw IOException in this case?

src/main/java/com/pinterest/secor/parser/QuboleClient.java
HenryCaiHaiying: And how about provide an option to skip qubole but still generate SUCCESS file?

src/main/java/com/pinterest/secor/parser/QuboleClient.java
shawnsnguyen: Actually, it's not necessary to return anything. If an exception occurs, the task will exit and the next finalization run will take care of finalizing any partitions that did not succeed.

src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -376,6 +376,8 @@ public String getHiveTableName(String topic) {
         return mProperties.getString(key, null);
     }
 
+    public boolean getSkipQuboleAddPartition() { return getBoolean("secor.skip.qubole.add.partition", false); }
```
HenryCaiHaiying: secor has the convention that config parameters are explicitly defined in secor.common.properties, usually no java code defaulting.

src/main/java/com/pinterest/secor/parser/PartitionFinalizer.java
```diff
@@ -168,7 +168,7 @@ private void finalizePartitionsUpTo(String topic, String[] uptoPartitions) throw
                             LOG.warn("HivePrefix is not defined.  Skip hive registration");
                         }
                     }
-                    if (hiveTableName != null) {
+                    if (hiveTableName != null && !mConfig.getSkipQuboleAddPartition()) {
```
HenryCaiHaiying: double negate might be a bit confusing.  It might be better to have a positive parameter name, e.g. enableQubole = true

src/main/java/com/pinterest/secor/parser/QuboleClient.java
```diff
@@ -31,6 +31,8 @@
  * @author Pawel Garbacki (pawel@pinterest.com)
  */
 public class QuboleClient {
+    private static final long MAX_QUBOLE_WAIT_TIME_MS = 300000; // 5 minutes
```
HenryCaiHaiying: This might be a candidate for the config parameter.

src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -498,6 +506,10 @@ public long getLong(String name) {
         return mProperties.getLong(name);
     }
 
+    public long getLong(String name, long defaultValue) {
```
HenryCaiHaiying: I think you can remove this method now.

PR #264, opened 2016-10-09
Protobuf 3 and google.protobuf.Timestamp type

prasincs:
Upgrading protobuf dependencies to 3.1.0 which is the latest stable. It is backwards compatible with the existing implementation that uses uint64 for timestamps. The type is checked at parsing time, and the relevant method is called to extract millis, so the rest of the system should continue to work as expected. 

If the tests look good, I can merge them into the same protobuf file to minimize confusion later.

### DIFF
--- a/src/main/java/com/pinterest/secor/parser/TimestampedMessageParser.java
+++ b/src/main/java/com/pinterest/secor/parser/TimestampedMessageParser.java
@@ -97,10 +97,6 @@ public abstract class TimestampedMessageParser extends MessageParser implements 
         return timestampMillis;
     }
 
-    protected static long toMillis(final Timestamp timestamp) {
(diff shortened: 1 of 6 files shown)


src/main/java/com/pinterest/secor/parser/TimestampedMessageParser.java
```diff
@@ -16,6 +16,8 @@
  */
 package com.pinterest.secor.parser;
 
+import com.google.protobuf.Timestamp;
```
HenryCaiHaiying: Timestamp is a protobuf specific class, should the new method 'toMillis' live in the  ProtobufMessageParser class?

HenryCaiHaiying:
Looks good.  Thanks for the contribution.

PR #265, opened 2016-10-13
Allow upload on oldest file per topic per partition

morrifeldman:
Add boolean configuration secor.file.age.youngest


HenryCaiHaiying:

PR #269, opened 2016-10-17
expose an extractTimestampMillis on just bytes for ProtobufMessageParser.

prasincs:
I use this class for verifying/indexing protobuf files by time after they've been uploaded in S3 (as Seqfiles, so the process is the same as processing JSON files). I'd send the rest of the project as a PR but it's not quite ready yet. This minor fix is to allow extracting timestamp on the underlying byte array and topic combination without changing the existing method signature that's used upstream and exposed by the service. I don't think adding this in TimestampedMessageParser is a good idea but I can see this being useful there too.


HenryCaiHaiying:
Looks OK to me.

PR #273, opened 2016-10-18
GCS daily and hourly folder structure support

ahsandar:
it allows to have folder structure for `GCS`  with date and hour partitions `Buckets/kafka-secor-backup/env/20161017/01/Logs/offset=0`

### DIFF
--- a/src/main/config/secor.common.properties
+++ b/src/main/config/secor.common.properties
@@ -110,7 +110,7 @@ secor.gs.path=data
 secor.gs.path.partition.daily=false
 secor.gs.path.partition.daily.format=yyyyMMdd
 secor.gs.path.partition.hourly=false
-secor.gs.path.partition.hourly=hh
+secor.gs.path.partition.hourly.format=hh
 
 # Use direct uploads
 # WARNING: disables resumable uploads, files are uploaded in a single request
--- a/src/main/config/secor.dev.gs.properties
+++ b/src/main/config/secor.dev.gs.properties
@@ -13,6 +13,12 @@ secor.gs.bucket=secor_gs
 # Google cloud storage path where files are stored within the bucket.
 secor.gs.path=data
 
+#partiotion uplaod in Google Clour Storage
+secor.gs.path.partition.daily=false
+secor.gs.path.partition.daily.format=yyyyMMdd
+secor.gs.path.partition.hourly=false
+secor.gs.path.partition.hourly.format=hh
+
 ################
 # END MUST SET #
 ################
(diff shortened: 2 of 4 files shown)


HenryCaiHaiying:
There is already hourly/minutely partition support in secor, e.g. through partitioner.granularity.hour, can you use that feature?

ahsandar:
@HenryCaiHaiying `partitioner.granularity.hour` enforces to write under hours and not to a dated folder if required and there was a PR opened related to having option to changing date format which wasn't merged because that could cause trouble with data import to Hive from S3 (can't find it right now).  so  it has a different functionality with felxibility for GCS uploads

ahsandar:
here it is https://github.com/pinterest/secor/pull/74, indicating limitations due to HIVE

HenryCaiHaiying:
When people use partitioner.granularity.hour for hourly partitioning, it's usually put in into the following folder structure:
   .../topc1/dt=2016-10-10/hr=01

And on hive table side, you define the hive table takes two partition column, e.g.:

create external table ... (
)
PARTITIONED BY ( 
  `dt` string, 
  `hr` string)

Is this what you want?  You can take a look at TimestampdMessageParser

Pawel didn't like customizing message parser class too much in the beginning, but we end up convincing him to support the hourly partition scheme.

ahsandar:
I m looking for the flexibility to have different date and hour format and support to only write to a dated folder with no hourly partitioning,  is that possible currently in secor?

HenryCaiHaiying:
The format in TimestampedMessageParser is currently hardcoded, you can
modify it to take in format config parameters.

I am not exactly sure what is a dated folder with no hourly partitioning
but still supporting hour format?

ahsandar:
option to write to both type of buckets 
`Buckets/kafka-secor-backup/env/20161017/Logs/offset=0`
`Buckets/kafka-secor-backup/env/20161017/01/Logs/offset=0`,
 will check the `TimestampedMessageParser` and have  a PR for it if that is more suitable

ahsandar:
from the looks of the code in `TimestampedMessageParser` it should always create a `dt=yyyy-MM-dd` partition, but I don't seem to be getting that, does enabling `secor.finalizer.lookback.periods=` initiates the partitions ?

HenryCaiHaiying:
Are you using one of the subclasses of TimestampedMessageParser, e.g.
jsonMessageParser or ProtobufMessageParser?

ahsandar:
This PR is outdated, alredy closed it as this was discarded, not sure what you mean by that

PR #275, opened 2016-10-21
Flexible format and DailyOffsetPartition Parser

ahsandar:
Made Timestamped MessageParser to have flexible prefix and partition format for date, hour and minute, also added daily offset partitioning message parser  to support following format to have partiotions backup up per day with hour and minute granularity 
/mnt/secor_data/message_logs/backup/28248_12/Log/20161021/02/of fset\=0/1_0_00000000000000025658.gz


ahsandar:
i have updated the access modifier on the fields , there is no need to `override` these as they are config settings that are already defined and in essence shouldn't be overridden during execution once defined., please review

HenryCaiHaiying:
Thanks for the contribution

PR #279, opened 2016-10-27
Partitioned parser with no offset folder 

ahsandar:
partitioned parser class that doesn't add an offset folder and flexible offset folder naming format defined in config.


src/main/java/com/pinterest/secor/parser/MessageParser.java
```diff
@@ -50,6 +52,10 @@ public MessageParser(SecorConfig config) {
         }
     }
 
+    static String usingOffsetPrefix(SecorConfig config) {
+        return config.getString("secor.offsets.prefix", "offset=");
```
HenryCaiHaiying: In secor, we keep the convention that the default value is specified in secor.common.properties, not in Java code.  You don't need the 2nd argument for default

HenryCaiHaiying:
Thanks.

PR #283, opened 2016-11-03
Extend ProgressMonitorMain to run in a loop.

liamstewart:
The interval between stats exports is set by
`monitoring.interval.seconds`. Setting to zero (the default) causes the
monitor to export once and then exit (existing behavior).


HenryCaiHaiying:
LGTM.  Thanks.

PR #284, opened 2016-11-03
Make consumer group prefix on statsd metrics optional.

liamstewart:
The default is the current behavior, which is to include the consumer
group as the overall prefix. Disable by setting
`secor.prefixWithConsumerGroup` to false.


HenryCaiHaiying:
Looks good to me.

PR #285, opened 2016-11-03
Allow auto.offset.reset to be specified in configuration.

liamstewart:
In most cases, the value should not be set to anything other than
"smallest". However, "largest" can be useful when transitioning from
to secor from another logsaver in situations where everything up to
largest has already been saved.

### DIFF
--- a/src/main/config/secor.common.properties
+++ b/src/main/config/secor.common.properties
@@ -128,6 +128,14 @@ secor.zookeeper.path=/
 # Impacts how frequently the upload logic is triggered if no messages are delivered.
 kafka.consumer.timeout.ms=10000
 
+# Where consumer should read from if no committed offset in zookeeper.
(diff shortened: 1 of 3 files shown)


liamstewart:
This duplicates #257. I didn't see that one until just now. We've had this change in place for a while locally - getting around to pushing upstream. Feel free to either pick this PR or close it.

HenryCaiHaiying:
Close this one since it's a duplicate

liamstewart:
But now both #257 and this one are closed, unless there's another duplicate that I haven't found?

HenryCaiHaiying:
I think I made a mistake, I should keep this one.

For some reason, I cannot reopen it anymore.  Can you try reopen or create another one?

liamstewart:
Re-opened as #286

PR #286, opened 2016-11-04
Allow auto.offset.reset to be specified in configuration.

liamstewart:
In most cases, the value should not be set to anything other than
"smallest". However, "largest" can be useful when transitioning from
to secor from another logsaver in situations where everything up to
largest has already been saved.


HenryCaiHaiying:
LGTM.

liamstewart:
Thanks for merging my PRs. Do you have an idea of when the next release of secor will be?

HenryCaiHaiying:
Currently no plan.  If you need one, I can cut one this week.

PR #291, opened 2016-11-29
Flexible Delimited File Reader Writer

ahsandar:
Need a flexible delimiter reader writer as new line delimiter adds an additional new line if data already is with new lines, which causes bad records for data to be imported in to Big Query from Google Cloud Storage


src/main/java/com/pinterest/secor/io/impl/FlexibleDelimitedFileReaderWriterFactory.java
```diff
@@ -0,0 +1,166 @@
...
+        FileSystem fs = FileUtil.getFileSystem(path.getLogFilePath());
+        this.mCountingStream = new CountingOutputStream(fs.create(fsPath));
+        this.mWriter = (codec == null) ? new BufferedOutputStream(
+        this.mCountingStream) : new BufferedOutputStream(
+        codec.createOutputStream(this.mCountingStream,
+        mCompressor = CodecPool.getCompressor(codec)));
+      }
+
+      public byte delimterWriter() {
+        byte delimiter = '\\';
```
HenryCaiHaiying: '\n'?

src/main/java/com/pinterest/secor/io/impl/FlexibleDelimitedFileReaderWriterFactory.java
```diff
@@ -0,0 +1,166 @@
...
+        @Override
+        public long getLength() throws IOException {
+          assert this.mCountingStream != null;
+          return this.mCountingStream.getCount();
+        }
+
+        @Override
+        public void write(KeyValue keyValue) throws IOException {
+          this.mWriter.write(keyValue.getValue());
+          if (DELIMITER != 92){
```
HenryCaiHaiying: don't use hard-coded 92

src/main/config/secor.common.properties
```diff
@@ -327,6 +327,8 @@ secor.file.extension=
 
 # The secor file reader/writer used to read/write the data, by default we write sequence files
 secor.file.reader.writer.factory=com.pinterest.secor.io.impl.SequenceFileReaderWriterFactory
+secor.file.reader.Delimiter=\n
```
HenryCaiHaiying: Do you need to escape newline char?

src/main/java/com/pinterest/secor/io/impl/FlexibleDelimitedFileReaderWriterFactory.java
HenryCaiHaiying: If you use charAt(0), you need to verify that input config only contains one char.

And are you able to read in newline char that way without escaping?

ahsandar: `\` is an escape character , thats why using it as it shouldn't be a delimiter,  as for the length of the delimiter, we don't need to identify the length as first character will become the delimiter for  value `\n`   that is as whole charAt(0) converted to `10`.

ahsandar: and if a long string is passed only the first character will be taken as a delimiter as a delimeter should be of length 1

src/main/java/com/pinterest/secor/io/impl/FlexibleDelimitedFileReaderWriterFactory.java
ahsandar: updated to use value converting to byte value for `\`

ahsandar: added comments to the config values to make it more clear to use

src/main/java/com/pinterest/secor/io/impl/FlexibleDelimitedFileReaderWriterFactory.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    public byte delimterReader() {
+      byte delimiter = '\n';
+      try {
+        String readerDelimiter = SecorConfig.load().getFileReaderDelimiter();
+        if (!readerDelimiter.isEmpty()){
+          delimiter = (byte)readerDelimiter.charAt(0);
+        }
+      }catch(ConfigurationException e){}
```
HenryCaiHaiying: Coding style: need a space between } and catch
Usually you should not ignore the Exception, it's better to throw it up or convert to a RuntimeException
The method name is better named as getReaderDelimter()

src/main/java/com/pinterest/secor/io/impl/FlexibleDelimitedFileReaderWriterFactory.java
```diff
@@ -0,0 +1,166 @@
...
+        return delimiter;
+      }
+
+      @Override
+      public KeyValue next() throws IOException {
+        ByteArrayOutputStream messageBuffer = new ByteArrayOutputStream();
+        int nextByte;
+        while ((nextByte = mReader.read()) != DELIMITER) {
```
HenryCaiHaiying: Do you need to handle the case that delimiter is part of the data (i.e. do you need to add escaping logic in case delimiter needs to be escaped?

src/main/java/com/pinterest/secor/io/impl/FlexibleDelimitedFileReaderWriterFactory.java
HenryCaiHaiying: You should prevent other developers misuse this parameter if you are only going to take char(0), the check is needed to prevent those misuses.

src/main/java/com/pinterest/secor/io/impl/FlexibleDelimitedFileReaderWriterFactory.java
```diff
@@ -0,0 +1,166 @@
...
+        @Override
+        public long getLength() throws IOException {
+          assert this.mCountingStream != null;
+          return this.mCountingStream.getCount();
+        }
+
+        @Override
+        public void write(KeyValue keyValue) throws IOException {
+          this.mWriter.write(keyValue.getValue());
+          if (DELIMITER != (byte)'\\'){
```
HenryCaiHaiying: The escaping logic is very confusing there.  Why not use empty writer delimiter as the signal to not write the delimiter to output?

src/main/java/com/pinterest/secor/io/impl/FlexibleDelimitedFileReaderWriterFactory.java
ahsandar: Delimiter can't be part of data otherwise no way to identify where new message start and where it ends, it needs to be unique, no need for that case

src/main/java/com/pinterest/secor/io/impl/FlexibleDelimitedFileReaderWriterFactory.java
ahsandar: `byte` value can't be for an empty `String` , and `\` is an escape character in creation of a string  so it can't be used on its own to be added between messages pulled out of secor as an   single message of `\` which would mean empty message with just `\`

src/main/java/com/pinterest/secor/io/impl/FlexibleDelimitedFileReaderWriterFactory.java
ahsandar: added check in secor config class to throw exception if length of delimiter is greater than 1

src/main/java/com/pinterest/secor/io/impl/FlexibleDelimitedFileReaderWriterFactory.java
ahsandar: updated throwing exception in catch block now

HenryCaiHaiying:
I think other changes are fine, but I don't like the way you are using \\ to tell whether there is write delimiter.  It's a hack.  I think it's better to use an explicit boolean variable to indicate that no write delimiter has been specified.

ahsandar:
@HenryCaiHaiying  added boolean logic to writeDelimiter, please review for merge

src/main/java/com/pinterest/secor/io/impl/FlexibleDelimitedFileReaderWriterFactory.java
```diff
@@ -122,6 +122,7 @@ public void close() throws IOException {
     private final BufferedOutputStream mWriter;
     private Compressor mCompressor = null;
     private byte DELIMITER = getWriterDelimiter();
```
HenryCaiHaiying: One minor comment, the instance variable name 'DELIMITER' should not be all uppercase.  All uppercase is reserved for constant.  I think you can either do 'mDelimiter' or 'delimiter'

ahsandar:
updated

ahsandar:
update: 	changed `\\` to `\n`

HenryCaiHaiying:
Thanks for your contribution and your patience.

ahsandar:
Thanks for your time

PR #292, opened 2016-11-30
Fix DelimitedTextFileWriter.getLength

kpavel:
Currently, at least for DelimitedTextFileWriter, the getLength returned value doesn't take into account data that not flushed yet (returns 0).
As a result, in Uploader.checkTopicPartition wrong decision can happen:  "size >= mConfig.getMaxFileSizeBytes()" while data already cashed and much larger than "secor.max.file.size.bytes".

It is worth to check other writers as well.


HenryCaiHaiying:
Looks good, thanks.

PR #298, opened 2017-01-20
Allow secor.properties to specify AWS S3 path style bucket specification

robmccoll:
This allows users to specify virtual host style (current way - stays by default if config not changed) and path style bucket specification for S3 in secor properties. See [S3 doc](http://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html).

I wanted to do some local testing with [Minio](https://www.minio.io/) which needed path style. Otherwise, Secor works perfectly with Minio using the S3 compatible interface :-)

### DIFF
--- a/src/main/config/secor.common.properties
+++ b/src/main/config/secor.common.properties
@@ -54,6 +54,10 @@ aws.proxy.http.port=
 aws.region=
 aws.endpoint=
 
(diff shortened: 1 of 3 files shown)


HenryCaiHaiying:
Thanks for the contribution

PR #299, opened 2017-01-23
if no prefix/topic, don't add empty component to path

robmccoll:
I had problems with paths ending up with an extra `/`  when `secor.s3.path` was empty which got encoded as a `%2f` in the path. This fixes that problem and potentially others associated with joining empty components into the path. I may just be doing something wrong elsewhere though?

### DIFF
--- a/src/main/java/com/pinterest/secor/common/LogFilePath.java
+++ b/src/main/java/com/pinterest/secor/common/LogFilePath.java
@@ -146,10 +146,10 @@ public class LogFilePath {
 
     public String getLogFileParentDir() {
         ArrayList<String> elements = new ArrayList<String>();
-        if(mPrefix.length() > 0) {
+        if (mPrefix != null && mPrefix.length() > 0) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/pinterest/secor/common/LogFilePath.java
```diff
@@ -146,8 +146,12 @@ public LogFilePath withPrefix(String prefix) {
 
     public String getLogFileParentDir() {
         ArrayList<String> elements = new ArrayList<String>();
-        elements.add(mPrefix);
-        elements.add(mTopic);
+        if(mPrefix.length() > 0) {
```
HenryCaiHaiying: If you want to be safe, you can do:
   if (mPrefix != null && mPrefix.length()) > 0 ) {

On the coding style side, there should be a space after 'if'

HenryCaiHaiying:
Looks good to me, thanks.

PR #304, opened 2017-02-07
Added configurable date prefixes

dnguyen0304:
The `DateMessageParser` provides flexibility through timestamp format strings, but then applies a hard-coded `dt=` prefix to the S3 path.  This configuration is actually already support through `partitioner.granularity.date.prefix` in `secor.common.properties` so I updated  `DateMessageParser.extractPartitions()` to leverage the existing functionality.

I also added `testDatePrefix()`, fixed existing tests that were now failing, and made minor formatting changes (e.g. indentation) to conform with the rest of the package.

### DIFF
--- a/src/main/java/com/pinterest/secor/parser/DateMessageParser.java
+++ b/src/main/java/com/pinterest/secor/parser/DateMessageParser.java
@@ -58,7 +58,7 @@ public class DateMessageParser extends MessageParser {
         inputFormatter.setTimeZone(timeZone);
         outputFormatter.setTimeZone(timeZone);
 
-        mDtPrefix = (TimestampedMessageParser.usingDatePrefix(config) == null) ? "" : TimestampedMessageParser.usingDatePrefix(config);
+        mDtPrefix = TimestampedMessageParser.usingDatePrefix(config);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/com/pinterest/secor/parser/DateMessageParser.java
```diff
@@ -39,20 +39,26 @@
...
     public DateMessageParser(SecorConfig config) {
         super(config);
+
         TimeZone timeZone = config.getTimeZone();
         inputPattern = mConfig.getMessageTimestampInputPattern();
         inputFormatter = new SimpleDateFormat(inputPattern.toString());
         inputFormatter.setTimeZone(timeZone);
         outputFormatter.setTimeZone(timeZone);
+
+        mDtPrefix = (TimestampedMessageParser.usingDatePrefix(config) == null) ? "" : TimestampedMessageParser.usingDatePrefix(config);
```
HenryCaiHaiying: TimestampedMessageParser.usingDatePrefix() already provides a default value, I think you can simply do:

    mDtPrefix = TimestampedMessageParser.usingDatePrefix(config)

HenryCaiHaiying:
Thanks for the contribution.

dnguyen0304:
Not a problem, Henry.  I'm a big fan of the project!  Thank _you_ for putting in the effort to maintain it.

Issue #305, opened 2017-02-10
JSON messages with RFC 3339 / ISO 8601 nanosecond timestamps not partitioned as expected to S3


gwilym:
When I analyse the files placed on S3 by secor I find there are messages being placed incorrectly with regards to date partitioning. Each JSON message has a top-level `timestamp` property which can be in either `2006-01-02T15:04:05.999999999Z` format or, occasionally, `2006-01-02T15:04:05Z` (as in, without nanoseconds, from systems which do not support such resolution).

Note: nanoseconds may not be a culprit here as the title suggests, that's just the majority of values in my dataset.

I've tried this with 0.22 from maven, and master (aee073b6a105f4365f50cabac1248747781a7a9a) from github.

It seems to work fine for the majority of events. However, when I scan the files for the dates of each message, I get results like...

```
# aws s3 ls --recursive s3://BUCKETNAME/secor-test-9/ | tr -s ' ' | cut -d' ' -f4 | sort | xargs -I{} sh -c 'echo {}; aws s3 cp s3://BUCKETNAME/{} - | zcat | jq ".timestamp[0:10]" | sort | uniq -c'
```

```
secor-test-9/athena-integration-1/dt=2017-01-31/1_22_00000000000002657819.gz
  19921 "2017-01-31"
      2 "2017-02-01"

secor-test-9/athena-integration-1/dt=2017-02-01/1_0_00000000000002675808.gz
  19302 "2017-02-01"
      3 "2017-02-02"
      1 "2017-02-03"

secor-test-9/athena-integration-1/dt=2017-02-01/1_11_00000000000002717065.gz
      2 "2017-02-02"
      2 "2017-02-03"
      5 "2017-02-04"
      3 "2017-02-05"
      2 "2017-02-06"

secor-test-9/athena-integration-1/dt=2017-02-01/1_12_00000000000002655329.gz
      3 "2017-01-31"
  33460 "2017-02-01"
      1 "2017-02-02"

secor-test-9/athena-integration-1/dt=2017-02-04/1_3_00000000000002720791.gz
      2 "2017-02-02"
  12869 "2017-02-04"
```

For one specific example above `secor-test-9/athena-integration-1/dt=2017-02-01/1_11_00000000000002717065.gz` the timestamps of each line are

```
"2017-02-02T22:33:03.12533900Z"
"2017-02-02T22:46:12.26030700Z"
"2017-02-03T00:49:04.27147800Z"
"2017-02-03T05:21:52.79127500Z"
[...]

PR #307, opened 2017-02-16
Let DateMessageParser use partitioner.granularity.date.format

gwilym:
The default 'yyyy-MM-dd' should match the existing behaviour of DateMessageParser.

I'm successfully using this to address #305 plus another side effect of having hive-compatible hourly partitions through DateMessageParser through the following configs:

```
secor.message.parser.class=com.pinterest.secor.parser.DateMessageParser
partitioner.granularity.date.prefix=
message.timestamp.input.pattern=yyyy-MM-dd'T'HH:mm:ss
partitioner.granularity.date.format='yr='yyyy'/mo='MM'/dy='dd'/hr='HH
```

Producing S3 paths like `secor/topicname/yr=2017/mo=02/dy=16/hr=00/1_0_00000000000484693804.gz`.


HenryCaiHaiying:

PR #308, opened 2017-02-17
Docker image for secor

enisher:
Build of secor docker image. Could be used as a base image or as a getting started point for newcomers.


HenryCaiHaiying:

PR #309, opened 2017-02-21
Add case for transforming micors to millis

serzh:
Just adding another case for transforming microseconds to milliseconds.


HenryCaiHaiying:

PR #310, opened 2017-02-22
Add SplitFieldMessageParser

DmitryBochorishvili:
Splits events into multiple output tables by a field value and partitions each table by date.

This pull request tries to address a use-case when Kafka topic contains events of multiple types with different schema and we would like to store events of each type as a separate table in the final storage. All events have common header with a discriminator field (event type) and timestamp.

Unfortunately secor doesn't seem to account for this use case. And I had to cut some corners to minimize changes to secor codebase:
 * the new parser doesn't support partition finalization
 * date format is hardcoded.  A composite like approach to plugin existing date parser would cause performance penalty since it leads to deserialization of json message two times.

I completely understand if you reject the pull request because of those shortcomings. I just don't see a nicer way to implement that use case within bounds of current secor design.
And any feedback would be much appreciated.

### DIFF
--- a/src/main/java/com/pinterest/secor/parser/SplitByFieldMessageParser.java
+++ b/src/main/java/com/pinterest/secor/parser/SplitByFieldMessageParser.java
@@ -1,0 +1,98 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.pinterest.secor.parser;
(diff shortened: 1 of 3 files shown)


src/main/java/com/pinterest/secor/parser/SplitByFieldMessageParser.java
```diff
@@ -0,0 +1,98 @@
...
+        String[] timestampPartitions = generatePartitions(timestampMillis, mUsingHourly, mUsingMinutely);
+        return ArrayUtils.addAll(new String[]{eventType}, timestampPartitions);
+    }
+
+    @Override
+    public String[] getFinalizedUptoPartitions(List<Message> lastMessages,
+                                               List<Message> committedMessages) throws Exception
+    {
+        // Partition finalization is not supported
+        return null;
```
HenryCaiHaiying: It seems to me you would still be able to implement this method.  You are just going to have 'eventType' as the first element in the partition list.

DmitryBochorishvili: Yes, it would work if we had a fixed and known set of eventTypes. Then `getPreviousPartitions` method would enumerate both though event types and date.

In my case the set of possible values of eventTypes is not known in advance and I don't see a way to enumerate them in `getPreviousPartitions` method. As the result partitions only for a single eventType will be finalized.
It sounds like a partially correct implementation of finalization logic and users may have hard time figuring out why some eventTypes got finalized and others not. What do you of it?

HenryCaiHaiying:
In that case, maybe just throw a UnSupportedOperationException, it's better than return null.

DmitryBochorishvili:
That's a good point to throw UnSupportedOperationException. Added a commit with the suggested change.

HenryCaiHaiying:

PR #312, opened 2017-03-01
added message parser option to docker image

ryansmithevans:


HenryCaiHaiying:
Looks good.

enisher:
@ryansmithevans glad to see you adding more options. 👍 
FYI you don't have to add mappings for all options. SECOR_EXTRA_OPTS env variable can be used instead. E.g:
```
SECOR_EXTRA_OPTS="-Dsecor.consumer.threads=14 -Dsecor.offsets.prefix=offset="
```
Supporting of all rarely used secor config options mapped in `docker-entrypoint` will not be easy to maintain.

PR #315, opened 2017-03-02
fix property name

ahsandar:
fixed property name, currently takes default so not configurable 
https://github.com/pinterest/secor/blob/master/src/main/config/secor.common.properties#L219


HenryCaiHaiying:
Change looks fine.  Can you retitle the PR, 'fox' --> 'fix'?

ahsandar:
@HenryCaiHaiying done

HenryCaiHaiying:
Thanks for the change.

PR #316, opened 2017-03-02
MetricCollector component for ingesting secor metrics into a monitoring service

enisher:
The component is pluggable, you can configure ingestion of metrics to any monitoring system by implementing MetricCollector interface.
By default metrics are sent to Ostrich, statistic library bundled into secor.

Metrics:
- consumer.message_size_bytes - records distribution of message sizes processed by secor (min/max/avg/count)
- consumer.throughput_bytes - total number of bytes processed by secor
- consumer.message_errors.count - number of errors during message parsing
- uploader.file_uploads.count - number of files uploaded

We are going to implement DataDogCollector for our set up. However, I tried to make the interface as abstract as possible, so potentially connector to any metric collection system can be implemented.

### DIFF
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -565,4 +565,8 @@ public class SecorConfig {
     public String getThriftProtocolClass() {
         return mProperties.getString("secor.thrift.protocol.class");
     }
+
+    public String getMetricsCollectorClass() {
+        return mProperties.getString("secor.monitoring.metrics.collector.class");
+    }
 }
--- a/src/main/java/com/pinterest/secor/consumer/Consumer.java
+++ b/src/main/java/com/pinterest/secor/consumer/Consumer.java
@@ -21,21 +21,19 @@ import com.pinterest.secor.common.OffsetTracker;
 import com.pinterest.secor.common.SecorConfig;
 import com.pinterest.secor.message.Message;
 import com.pinterest.secor.message.ParsedMessage;
+import com.pinterest.secor.monitoring.MetricCollector;
 import com.pinterest.secor.parser.MessageParser;
-import com.pinterest.secor.transformer.MessageTransformer;
-import com.pinterest.secor.uploader.Uploader;
-import com.pinterest.secor.uploader.UploadManager;
 import com.pinterest.secor.reader.MessageReader;
+import com.pinterest.secor.transformer.MessageTransformer;
+import com.pinterest.secor.uploader.UploadManager;
+import com.pinterest.secor.uploader.Uploader;
 import com.pinterest.secor.util.ReflectionUtil;
 import com.pinterest.secor.writer.MessageWriter;
-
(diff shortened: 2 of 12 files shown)


enisher:
@HenryCaiHaiying, any feedback on this?

I want to expose even more metrics to have a better picture of what's going on inside of the system.
Do you think this is useful for secor users?
Do you think the approach is correct?

src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -21,11 +21,7 @@
 import org.apache.commons.configuration.PropertiesConfiguration;
 import org.apache.commons.lang.StringUtils;
 
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.Map;
-import java.util.Properties;
-import java.util.TimeZone;
+import java.util.*;
```
HenryCaiHaiying: Don't use '*', each class needs to be individually imported.

src/main/java/com/pinterest/secor/consumer/Consumer.java
```diff
@@ -145,6 +146,8 @@ protected boolean consumeNextMessage() {
                 final double DECAY = 0.999;
                 mUnparsableMessages *= DECAY;
             } catch (Throwable e) {
+                mMetricCollector.increment("consumer.message_errors.count", "topic:" + rawMessage.getTopic());
```
HenryCaiHaiying: Can we make 'topic:' as an explicit argument to 'increment' call instead of concat string

src/main/java/com/pinterest/secor/monitoring/MetricCollector.java
```diff
@@ -0,0 +1,46 @@
...
+ * All methods should be non-blocking and do not throw exceptions.
+ */
+public interface MetricCollector {
+    /**
+     * Convenience method equivalent to {@link #increment(String, int, String...)}.
+     *
+     * @param label metric name
+     * @param tags array of tags to be added to the data
+     */
+    void increment(String label, String... tags);
```
HenryCaiHaiying: Does this mean the increment amount is 1?

src/main/java/com/pinterest/secor/common/SecorConfig.java
enisher: Fixed. Sorry, didn't pay attention to the imports.

src/main/java/com/pinterest/secor/consumer/Consumer.java
enisher: Done

src/main/java/com/pinterest/secor/monitoring/MetricCollector.java
enisher: Changed JavaDoc to declare that explicitly.

enisher:
@HenryCaiHaiying made changes according to your suggestions.

HenryCaiHaiying:
Thanks for the contribution.

Issue #328, opened 2017-04-12
Memory Footprint of Parquet Writers


thegeebe:
I'm getting `java.lang.OutOfMemoryError: Java heap space` errors when using the `ProtobufParquetFileReaderWriter`. I know a simple solution is usually to increase memory size, but the amount of data coming through wouldn't warrant it.

Some stats:
- Running as a StatefulSet on Kubernetes on AWS
- Kafka 0.10
- 120 Partitions in Kafka input topic. The Kafka partitions are hashed on the same keys that I want to partition my Secor output with, to ensure that no two threads need to write to the same s3 dir/minimize the number of local files each thread is required to write to.
- 5 Secor processes and 7 consumer threads.
- Each process runs in a docker container with 2.5G memory
- Each process has the java heap specified as 2G
- Each process gets 250m cpu
- Input data is Protobuf

This configuration is perfectly stable when using something simple like the `OffsetMessageParser` or even `ProtobufMessageParser`. However, when rolling my own parser that uses a field whose values are uniformly distributed across 577 unique integers, I very quickly run out of memory. The size of my input data is probably < 10MB at the time the consumers begin to fail. If I increase the amount of CPU allocated to each process, it fails faster, presumably because it is able to process more events.

I don't think this is a protobuf issue, because other parsers/output groupers work fine. My initial reaction was that this could be the overhead of having 577 parquet writers, and that converting to parquet is more memory intensive than, say, a sequence file. However, each thread should only have a handful of writers if Secor is distributing the load the way I think it should be, considering how my topics are partitioned.

Currently I'm trying to connect a remote debugger to analyze the heap, but was wondering if you had any tips/insights into the memory overhead of the parquet writer.

---
This is my parser, which I don't believe is responsible for any sort of bad memory mangement:
```java
public class EventParser extends MessageParser {

	private static final Logger LOG = LoggerFactory.getLogger(EventParser.class);

	private EventConfig eventConfig;
	private ProtobufUtil protobufUtil;

	public EventParser(SecorConfig config) {
		super(config);

		eventConfig = new EventConfig(config);
		protobufUtil = new ProtobufUtil(config);
	}

	@Override
	public String[] extractPartitions(Message message) throws Exception {
		com.google.protobuf.Message decodedMessage = protobufUtil.decodeMessage(message.getTopic(), message.getPayload());

		String event_date = "event_date=" + decodedMessage.getField(decodedMessage.getDescriptorForType().findFieldByName("event_date")).toString();
[...]

PR #329, opened 2017-04-19
Bump Parquet and Expose Configuration Variables

thegeebe:
See discussion at #328 

Exposes row group size, page size, dictionary encoding, and validation as configurable options to pass to the parquet writer. Bumps Parquet to 1.9.0 which implements more memory efficient dictionary encoding, among other enhancements.

### DIFF
--- a/src/main/config/secor.common.properties
+++ b/src/main/config/secor.common.properties
@@ -381,3 +381,20 @@ secor.file.age.youngest=true
 # Class that manages metric collection.
 # Sending metrics to Ostrich is the default implementation.
 secor.monitoring.metrics.collector.class=com.pinterest.secor.monitoring.OstrichMetricCollector
+
+# Row group size in bytes for Parquet writers. Specifies how much data will be buffered in memory before flushing a
+# block to disk. Larger values allow for larger column chinks which makes it possible to do larger sequential IO.
+# Should be aligned with HDFS blocks. Defaults to 128MB in Parquet 1.9.
+parquet.block.size=134217728
+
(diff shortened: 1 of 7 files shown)


HenryCaiHaiying:
Looks good overall.  Can you add the 4 new parameters into secor.common.properties with some docs.

thegeebe:
Alright, I put in some comments and made the default values in the config the default values from ParquetWriter implementation; not sure what your convention is for params that have defaults defined in the code. Also I squashed the commits for cleanliness

src/main/config/secor.common.properties
```diff
@@ -381,3 +381,20 @@ secor.file.age.youngest=true
...
+# fine grained reading but have higher space overhead. Defaults to 1MB in Parquet 1.9.
+parquet.page.size=1048576
+
+# Enable or disable dictionary encoding for Parquet writers. The dictionary encoding builds a dictionary of values
+# encountered in a given column. Defaults to true in Parquet 1.9.
+parquet.enable.dictionary=true
+
+# Enable or disable validation for Parquet writers. Validates records written against the schema. Defaults to false in
+# Parquet 1.9.
+parquet.validation=false
```
HenryCaiHaiying: Looks good to me except there is one extra character at the end of line.

thegeebe: it's not a new character, just a notification that there is no newline. I added a new line and squashed

HenryCaiHaiying:
Thanks for the investigation and fix, this is a great contribution.

PR #338, opened 2017-05-17
Added timestamp in Message class and populating it while extracting o…

tygrash:
…ther fields from kafka message

@HenryCaiHaiying please review.

### DIFF
--- a/src/main/java/com/pinterest/secor/common/KafkaClient.java
+++ b/src/main/java/com/pinterest/secor/common/KafkaClient.java
@@ -141,8 +141,11 @@ public class KafkaClient {
             payloadBytes = new byte[payload.limit()];
             payload.get(payloadBytes);
         }
+        Long timestamp = null;
+        timestamp = messageAndOffset.message().timestamp();
+
         return new Message(topicPartition.getTopic(), topicPartition.getPartition(),
-                messageAndOffset.offset(), keyBytes, payloadBytes);
+                messageAndOffset.offset(), keyBytes, payloadBytes, timestamp);
     }
 
     private SimpleConsumer createConsumer(String host, int port, String clientName) {
--- a/src/main/java/com/pinterest/secor/common/KafkaClient.java
+++ b/src/main/java/com/pinterest/secor/common/KafkaClient.java
@@ -142,8 +142,9 @@ public class KafkaClient {
             payload.get(payloadBytes);
         }
         Long timestamp = null;
-        timestamp = messageAndOffset.message().timestamp();
-
+        if (mConfig.getUseKafkaTimestamp()) {
+            timestamp = messageAndOffset.message().timestamp();
+        }
         return new Message(topicPartition.getTopic(), topicPartition.getPartition(),
                 messageAndOffset.offset(), keyBytes, payloadBytes, timestamp);
     }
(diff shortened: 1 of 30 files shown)


src/main/java/com/pinterest/secor/common/KafkaClient.java
```diff
@@ -141,8 +141,11 @@ private Message getMessage(TopicPartition topicPartition, long offset,
             payloadBytes = new byte[payload.limit()];
             payload.get(payloadBytes);
         }
+        Long timestamp = null;
+        timestamp = messageAndOffset.message().timestamp();
```
HenryCaiHaiying: There will be a back compatibility issue here, either because:
1. People still use kafka 8 library
2. Or people on kafka 10 but don't want to use the timestamp from kafka message (they still want to retrieve the timestamp from other fields).

I think we need to introduce a config param, something like useKafkaTimestamp, and it should be false by default.

HenryCaiHaiying:
Left a comment for a backward compatibility issue.

src/main/java/com/pinterest/secor/common/KafkaClient.java
tygrash: Correct Henry, I have added the param `kafka.useKafkaTimestamp` in secor.common.properties and accordingly extracting timestamp, otherwise passing it as null. Please review and let me know if further changes required! Thank You.

src/main/java/com/pinterest/secor/parser/MessageParser.java
```diff
@@ -60,7 +60,7 @@ public ParsedMessage parse(Message message) throws Exception {
         String[] partitions = extractPartitions(message);
```
HenryCaiHaiying: In the old way, the extractPartitions(message) will extract the timestamp from the message and generate the partition array from that timestamp.

How does your newly extracted kafka timestamp used in that situation then?

tygrash: Added code for timestamp extraction based on configuration value and using it in partition generation.
Also added tests for the same in different parsers.

HenryCaiHaiying:
This change seems too big, changed too many subclasses.

How about you only change TimestampedMessageParser.java (most of other classes are subclasses of it).  Provide a new 'getTimestampMillis' method which provides a default implementation of:
    if (config.useKafkaMessage()) {
           return message.getTimestamp and rounded to millis
    } else {
           return extractTimestampMillis();
    }

And change the caller to call this new method instead.

tygrash:
Nice idea! Implemented the desired behaviour in TimestampedMessageParser and every subclass will call this default implementation.

src/main/java/com/pinterest/secor/parser/TimestampedMessageParser.java
```diff
@@ -126,6 +126,10 @@ static String usingMinutePrefix(SecorConfig config) {
         return config.getString("partitioner.granularity.minute.prefix", "min=");
     }
 
+    static boolean useKafkaTimestamp(SecorConfig config) {
+        return config.getBoolean("kafka.useTimestamp", false);
```
HenryCaiHaiying: This should call config.useKafkaTimestamp()

tygrash: Resolved.

HenryCaiHaiying:
Looks great, thanks for your patience.  A great contribution.

tygrash:
All thanks to you Henry. This is my first contribution to an open-source project and couldn't have been possible without your guidance. Really appreciate your help in all the reviews and bearing patience with me :)

Issue #342, opened 2017-06-08
Backing up Kafka messages timestamp


nachomdo:
Hi, 
As you are probably aware, Kafka 0.10 message format includes a timestamp field. I wonder if you are thinking about providing some mechanism to back up this field as well. At the moment, only Offset, Key and Payload fields are saved. 

I have made some changes for MessagePackSequenceFileReader and Writer in my fork to support this so, I could open a PR if this is something you want to have as part of Secor. 

Thanks 😃

HenryCaiHaiying:
In https://github.com/pinterest/secor/commit/cff776388fc497d4193d910feb182eeffeab74ab, tygrash added the capability of using the timestamp field from kafka 10 message.

How do you use the the timestamp in your fork?

nachomdo:
Thanks for replying so quickly. Maybe I'm missing something. What I want to do is backup messages from one environment including the timestamp and restoring them in other environment using a Kafka Client 0.10 that allows to specify the timestamp of a given message. 

The tool I have to restore messages is using a reader created with `MessagePackSequenceFileReaderWriterFactory` to read the files produced by Secor. The reader allows us to iterate over a collection of `KeyValue` which contain the fields: offset, key, value. I have noticed that, at least in the MessagePackSequenceWriter, the offset is encoded as part of the key. I have just leveraged the fact that `ParsedMessage` contains already the timestamp field, added a timestamp field to `KeyValue` and changed how the message is encoded and decoded using MessagePack to consider the new timestamp field as well. 

```java
        @Override
        public void write(KeyValue keyValue) throws IOException {
            byte[] kafkaKey = keyValue.getKafkaKey();
            // output size estimate
            // 1 - map header
            // 1 - message pack key
            // 9 - max kafka offset
            // 1 - message pack key
            // 9 - kafka timestamp
            // 1 - message pack key
            // 5 - max (sane) kafka key size
            // N - size of kafka key
            // = 26 + N
            ByteArrayOutputStream out = new ByteArrayOutputStream(26 + kafkaKey.length);
            MessagePacker packer = MessagePack.newDefaultPacker(out)
                    .packMapHeader((kafkaKey.length == 0) ? 2 : 3)
                    .packInt(KAFKA_MESSAGE_OFFSET)
                    .packLong(keyValue.getOffset())
                    .packInt(KAFKA_MESSAGE_TIMESTAMP)
                    .packLong(keyValue.getTimestamp());
            if (kafkaKey.length != 0) {
[...]

PR #344, opened 2017-06-12
Stores Kafka message timestamp as part of the encoded information

nachomdo:
This PR allows to store Kafka messages timestamp information as part of the encoded MessagePack record. 

This has been discussed here https://github.com/pinterest/secor/issues/342

### DIFF
--- a/src/main/java/com/pinterest/secor/io/KeyValue.java
+++ b/src/main/java/com/pinterest/secor/io/KeyValue.java
@@ -19,21 +19,23 @@ package com.pinterest.secor.io;
 /**
  * Generic Object used to read next message from various file reader
  * implementations
- * 
+ *
  * @author Praveen Murugesan (praveen@uber.com)
  *
  */
 public class KeyValue {
-	
+
 	private final long mOffset;
 	private final byte[] mKafkaKey;
 	private final byte[] mValue;
+	private final long mTimestamp;
 
 	// constructor
 	public KeyValue(long offset, byte[] value) {
 		this.mOffset = offset;
 		this.mKafkaKey = new byte[0];
 		this.mValue = value;
+		this.mTimestamp = -1;
 	}
 
 	// constructor
@@ -41,6 +43,15 @@ public class KeyValue {
 		this.mOffset = offset;
 		this.mKafkaKey = kafkaKey;
 		this.mValue = value;
+		this.mTimestamp = -1;
+	}
+
+	// constructor
+	public KeyValue(long offset, byte[] kafkaKey, byte[] value, long timestamp) {
+		this.mOffset = offset;
+		this.mKafkaKey = kafkaKey;
+		this.mValue = value;
+		this.mTimestamp = timestamp;
 	}
 
 	public long getOffset() {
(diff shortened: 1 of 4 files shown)


src/main/java/com/pinterest/secor/io/KeyValue.java
```diff
@@ -50,9 +61,20 @@ public long getOffset() {
...
 	public byte[] getValue() {
 		return this.mValue;
 	}
 
+	public long getTimestamp() {
+		return this.mTimestamp;
+	}
+
+	public boolean hasKafkaKey() {
+		return this.mKafkaKey.length != 0;
```
HenryCaiHaiying: Probably should check both null and zero length

src/main/java/com/pinterest/secor/io/impl/MessagePackSequenceFileReaderWriterFactory.java
```diff
@@ -76,13 +77,17 @@ public KeyValue next() throws IOException {
                 MessageUnpacker unpacker = MessagePack.newDefaultUnpacker(mKey.getBytes());
                 int mapSize = unpacker.unpackMapHeader();
                 long offset = 0;
+                long timestamp = 0;
```
HenryCaiHaiying: Should the default value of timestamp be -1, to be consistent with the KeyValue class?

src/main/java/com/pinterest/secor/io/impl/MessagePackSequenceFileReaderWriterFactory.java
```diff
@@ -161,5 +175,17 @@ public void write(KeyValue keyValue) throws IOException {
...
             this.mWriter.close();
         }
+
+        private int numberOfFieldsMappedInHeader(KeyValue keyValue) {
+            int fields = 1;
+
+            if (keyValue.hasKafkaKey())
+                fields++;
+
+            if (keyValue.hasTimestamp())
```
HenryCaiHaiying: What if keyValue does not have kafkaKey but has timestamp, would the field index correct in that case?

nachomdo:
Thanks @HenryCaiHaiying for taking the time to review these changes. I have addressed your comments and added a new test case for the scenario you were describing: a record with no key but with timestamp.

src/main/java/com/pinterest/secor/io/impl/MessagePackSequenceFileReaderWriterFactory.java
HenryCaiHaiying: How was this case addressed in your new fix?

nachomdo: I've added a test to check the scenario you are describing (a record with timestamp but no key) 
```
KeyValue kv4 =  (new KeyValue(23235, null, new byte[]{23, 45, 40 ,10, 122}, 1496318250));
```
This method is returning the number of fields that we are encoding together, not the field index. So for a record with offset but no key and no timestamp should return 1. If we had key, offset and timestamp it should return 3 and in case either the key or the timestamp is missing it will return 2.

HenryCaiHaiying: I see what you mean now, if the method only care about the number of fields, it should be fine.

HenryCaiHaiying:
Thanks for the contribution.

PR #352, opened 2017-07-08
Rebuild Iso8601MessageParser on top of TimestampedMessageParser

proger:
This makes it work with `partitioner.granularity.hour` and the like. While here, mentioned it in the README so it doesn't feel neglected.


HenryCaiHaiying:
Looks good, thanks.

PR #353, opened 2017-07-11
Fix for backward compatibility in case of kafkaTimestamp usage

tygrash:
Basically, idea is to use a class name from properties file and creating dynamic timestamp classes for 0.8 and 0.10 timestamps. This will cut out the timestamp factory dependency on excluded class in 0.8-dev build profile  and will use the configuration in run-time only, so that code is compiled and people can use the jar.


pom.xml
```diff
@@ -602,6 +610,20 @@
...
             <id>kafka-0.8-dev</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-compiler-plugin</artifactId>
+                        <version>2.0.2</version>
+                        <configuration>
+                            <excludes>
+                                <exclude>com/pinterest/secor/timestamp/Kafka10MessageTimestamp.java</exclude>
```
HenryCaiHaiying: If 'exclude' tag works, would you be able to exclude Kafka8MessageTimestamp.java in kafka-0.10-dev build profile?  This way you don't have to introduce a dynamic class, everything is static, each build profile bundles a different set of java classes.

tygrash: The only problem with that approach is KafkaMessageTimestampFactory.java, because it will call either Kafka10MessageTimestamp.java or Kafka8MessageTimestamp.java depending on the config, and since one of the class will not be available for compile(because we have excluded it), so `cannot find symbol` error starts coming during compile. Thats why I found this workaround.

src/main/config/secor.common.properties
```diff
@@ -181,10 +181,17 @@ kafka.dual.commit.enabled=true
...
+# Parameter which tells whether to extract Kafka message timestamp. This value is to be chose in case of 0.10.x kafka brokers.
+# Default value is false. Also specify `kafka.message.timestamp.className` as `com.pinterest.secor.timestamp.Kafka10MessageTimestamp`,
+# in case you are enabling this parameter as `true`.
 kafka.useTimestamp=false
 
+# Classname for the timestamp field you want to use.
+# Basically, it will be `Kafka8MessageTimestamp` for 0.8 kafka and `Kafka10MessageTimestamp` for 0.10 kafka,
+# Fully classified names are `com.pinterest.secor.timestamp.Kafka8MessageTimestamp` and
+# `com.pinterest.secor.timestamp.Kafka10MessageTimestamp`.
+kafka.message.timestamp.className=com.pinterest.secor.timestamp.Kafka8MessageTimestamp
```
HenryCaiHaiying: I think the default value should be Kafka10MessageTimestamp since the default build profile is now kafka 10.

src/main/java/com/pinterest/secor/common/KafkaClient.java
```diff
@@ -143,7 +148,7 @@ private Message getMessage(TopicPartition topicPartition, long offset,
         }
         Long timestamp = null;
         if (mConfig.useKafkaTimestamp()) {
-            timestamp = messageAndOffset.message().timestamp();
+            timestamp = mKafkaMessageTimestampFactory.create(mKafkaTimestampClassName).getTimestamp(messageAndOffset);
```
HenryCaiHaiying: Instead of calling factory.create(className) each time, can the factory have the timestampClass created in its constructor so you just need to get that class each time?  Use reflection to create the java class each time is expensive.

src/main/java/com/pinterest/secor/timestamp/KafkaMessageTimestampFactory.java
```diff
@@ -0,0 +1,19 @@
...
+public class KafkaMessageTimestampFactory {
+    private static final Logger LOG = LoggerFactory.getLogger(KafkaMessageTimestampFactory.class);
+
+    public static KafkaMessageTimestamp create(String kafkaTimestampClassName) {
+        KafkaMessageTimestamp kafkaMessageTimestamp = null;
+        try {
+            Class timestampClass = Class.forName(kafkaTimestampClassName);
+            kafkaMessageTimestamp = KafkaMessageTimestamp.class.cast(timestampClass.newInstance());
+        } catch (ClassNotFoundException | InstantiationException | IllegalAccessException e) {
+            LOG.error("Unable to create KafkaTimestampFactory", e);
```
HenryCaiHaiying: Just throw RuntimeException to stop the execution in this situation.

src/main/java/com/pinterest/secor/reader/MessageReader.java
```diff
@@ -160,9 +165,14 @@ public Message read() {
             RateLimitUtil.acquire(mCheckMessagesPerSecond);
         }
         MessageAndMetadata<byte[], byte[]> kafkaMessage = mIterator.next();
+
+        Long timestamp = null;
+        if (mConfig.useKafkaTimestamp()) {
+            timestamp = mKafkaMessageTimestampFactory.create(mKafkaMessageTimestampClassName).getTimestamp(kafkaMessage);
```
HenryCaiHaiying: Avoid calling reflection class creation each time.

src/main/java/com/pinterest/secor/timestamp/Kafka8MessageTimestamp.java
```diff
@@ -0,0 +1,17 @@
+package com.pinterest.secor.timestamp;
+
+import kafka.message.MessageAndMetadata;
+import kafka.message.MessageAndOffset;
+
+public class Kafka8MessageTimestamp implements KafkaMessageTimestamp {
+
+    @Override
+    public Long getTimestamp(MessageAndMetadata<byte[], byte[]> kafkaMessage) {
+        return null;
```
HenryCaiHaiying: If return null, how does caller deal with that situation?

tygrash:
Made the changes @HenryCaiHaiying 
Please review again.

src/main/java/com/pinterest/secor/timestamp/Kafka8MessageTimestamp.java
```diff
@@ -7,11 +7,11 @@
...
     @Override
     public Long getTimestamp(MessageAndMetadata<byte[], byte[]> kafkaMessage) {
-        return null;
+        return 0l;
     }
 
     @Override
     public Long getTimestamp(MessageAndOffset messageAndOffset) {
-        return null;
+        return 0l;
```
HenryCaiHaiying: Looks like both k8 and k10 will return a valid long, should the method now return a small long instead of big Long?

tygrash: Oh yes @HenryCaiHaiying, fixing this.I should have fixed it in previous commit only, my bad. Fixing it.

src/main/java/com/pinterest/secor/common/KafkaClient.java
```diff
@@ -141,10 +144,7 @@ private Message getMessage(TopicPartition topicPartition, long offset,
             payloadBytes = new byte[payload.limit()];
             payload.get(payloadBytes);
         }
-        Long timestamp = null;
```
HenryCaiHaiying: This logic 'if (mConfig.useKafkaTimestamp()) {' seems gone in the new code

src/main/java/com/pinterest/secor/common/KafkaClient.java
```diff
@@ -141,10 +144,7 @@ private Message getMessage(TopicPartition topicPartition, long offset,
...
-        if (mConfig.useKafkaTimestamp()) {
-            timestamp = messageAndOffset.message().timestamp();
-        }
+        Long timestamp = mKafkaMessageTimestampFactory.getKafkaMessageTimestamp().getTimestamp(messageAndOffset);
```
HenryCaiHaiying: Are you still using Long instead of long here.

src/main/java/com/pinterest/secor/common/KafkaClient.java
tygrash: long timestamp = 0l;
        if (mConfig.useKafkaTimestamp()) {
            timestamp = mKafkaMessageTimestampFactory.getKafkaMessageTimestamp().getTimestamp(kafkaMessage);
        }

Is this fine?

src/main/java/com/pinterest/secor/io/KeyValue.java
```diff
@@ -47,7 +47,7 @@ public KeyValue(long offset, byte[] kafkaKey, byte[] value) {
 	}
 
 	// constructor
-	public KeyValue(long offset, byte[] kafkaKey, byte[] value, long timestamp) {
+	public KeyValue(long offset, byte[] kafkaKey, byte[] value, Long timestamp) {
```
HenryCaiHaiying: should this be 'long'?

tygrash: yupp, it should be long only. Resolved!

HenryCaiHaiying:
Thanks for the change and your patience.

Issue #189, opened 2016-03-17
ProgressMonitor fails in case of topic inactivity


hibikir:
If a topic has been silent for the whole Kafka retention duration, Progress Monitor blows up completely, as it attempts to get an offset that is not available anymore, and responds with a plain Runtime Exception that is never caught:

```
java.lang.RuntimeException: Error fetching offset data. Reason: 1
    at com.pinterest.secor.common.KafkaClient.getMessage(KafkaClient.java:128)
    at com.pinterest.secor.common.KafkaClient.getCommittedMessage(KafkaClient.java:207)
    at com.pinterest.secor.tools.ProgressMonitor.getStats(ProgressMonitor.java:183)
    at com.pinterest.secor.tools.ProgressMonitor.exportStats(ProgressMonitor.java:131)
    at com.pinterest.secor.main.ProgressMonitorMain.main(ProgressMonitorMain.java:43)
```

At the very least, I think the exception should be caught, so that we still get information from other topics. My plan is to detect this specific error code, and then check if there are messages outstanding. If there aren't any, then our delay is zero (the topic is just very quiet). If there are some, our delay is the difference between the first and the last message, and throw a message into the log saying that it's possible that we have lost messages.

I'd love to hear if this is a sound plan, to hear if anyone has any better ideas, and to know if this enhancement is something you'd be interested in me PRing into the Pinterest repo.

HenryCaiHaiying:
Agree we definitely should handle this exception.  If you want to fix
further, notice that even though the message is no longer in the broker,
but you have the two message offsets (last write message from the producer
and last read message from the consumer), if they are the same then no lag,
otherwise you might print the offset difference *
heuristic-of-per-message-gap.

mcs1408:
Has there been any updates on this topic? I am currently having the same issue and would love to know how you fixed the problem.

HenryCaiHaiying:
Not yet, you are welcome to make some code contributions along this line.

mcs1408:
[...]

PR #356, opened 2017-07-28
Handle empty topics for the ProgressMonitor

Fluxx:
If a `TopicPartition` has had all of its messages on the broker compacted away, when you call `KafakaClient.getCommittedMessage` with that `TopicPartition` a `RuntimeException` was raised. This is due to the obvious fact that `getCommittedMessage` is attempting to return a `Message`, however there is no `Message` to return from the broker.

This is not, I would argue, an _exceptional_ situation. If the topic has low volume or an aggressive compaction configuration, often times the log on the broker will be empty. This generally does not cause problems, except in the case of #189, where the `ProgressMonitor` will crash on the error.

This is not ideal, as actual progress is totally caught up. So this change attempts to address this situation through two changes:

1. Change `KafakaClient.getCommittedMessage` to return `null` if there is no message found.
2. Update `ProgressMonitor.getStats` to record a `0` lag value if no committed message was found.

I've left some self-review notes below please take a look.

I'll also add that I didn't see any existing tests for the `ProgressMonitor` or `KafakaClient` so I did not add any. Let me know if you think it would be worth adding some however.

### DIFF
--- a/src/main/java/com/pinterest/secor/common/KafkaClient.java
+++ b/src/main/java/com/pinterest/secor/common/KafkaClient.java
@@ -31,6 +31,7 @@ import kafka.javaapi.TopicMetadataRequest;
 import kafka.javaapi.TopicMetadataResponse;
 import kafka.javaapi.consumer.SimpleConsumer;
 import kafka.message.MessageAndOffset;
+import org.apache.kafka.common.protocol.Errors;
 import org.apache.thrift.TException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -131,7 +132,7 @@ public class KafkaClient {
             consumer.close();
             int errorCode = response.errorCode(topicPartition.getTopic(), topicPartition.getPartition());
 
-            if (errorCode == 1) {
+            if (errorCode == Errors.OFFSET_OUT_OF_RANGE.code()) {
               throw new MessageDoesNotExistException();
             } else {
               throw new RuntimeException("Error fetching offset data. Reason: " + errorCode);
--- a/src/main/java/com/pinterest/secor/common/KafkaClient.java
+++ b/src/main/java/com/pinterest/secor/common/KafkaClient.java
@@ -59,6 +59,8 @@ public class KafkaClient {
         mKafkaMessageTimestampFactory = new KafkaMessageTimestampFactory(mConfig.getKafkaMessageTimestampClass());
     }
 
+    public class MessageDoesNotExistException extends RuntimeException {}
+
     private HostAndPort findLeader(TopicPartition topicPartition) {
         SimpleConsumer consumer = null;
         try {
@@ -127,8 +129,13 @@ public class KafkaClient {
         FetchResponse response = consumer.fetch(request);
         if (response.hasError()) {
             consumer.close();
-            throw new RuntimeException("Error fetching offset data. Reason: " +
-                    response.errorCode(topicPartition.getTopic(), topicPartition.getPartition()));
+            int errorCode = response.errorCode(topicPartition.getTopic(), topicPartition.getPartition());
+
+            if (errorCode == 1) {
+              throw new MessageDoesNotExistException();
+            } else {
+              throw new RuntimeException("Error fetching offset data. Reason: " + errorCode);
+            }
         }
         MessageAndOffset messageAndOffset = response.messageSet(
                 topicPartition.getTopic(), topicPartition.getPartition()).iterator().next();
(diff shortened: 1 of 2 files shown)


src/main/java/com/pinterest/secor/common/KafkaClient.java
```diff
@@ -127,8 +129,13 @@ private Message getMessage(TopicPartition topicPartition, long offset,
         FetchResponse response = consumer.fetch(request);
         if (response.hasError()) {
             consumer.close();
-            throw new RuntimeException("Error fetching offset data. Reason: " +
-                    response.errorCode(topicPartition.getTopic(), topicPartition.getPartition()));
+            int errorCode = response.errorCode(topicPartition.getTopic(), topicPartition.getPartition());
+
+            if (errorCode == 1) {
```
Fluxx: See response codes from the Kafka docs:

https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-PossibleErrorCodes.1

src/main/java/com/pinterest/secor/common/KafkaClient.java
```diff
@@ -212,6 +219,16 @@ public Message getCommittedMessage(TopicPartition topicPartition) throws Excepti
...
+        } catch (MessageDoesNotExistException e) {
+          // If a RuntimeEMessageDoesNotExistException exception is raised,
+          // the message at the last comitted offset does not exist in Kafka.
+          // This is usually due to the message being compacted away by the
+          // Kafka log compaction process.
+          //
+          // That is no an exceptional situation - in fact it can be normal if
+          // the topic being consumed by Secor has a low volume. So in that
+          // case, simply return null
+          return null;
```
Fluxx: I also think a suitable alternative to catching this error here is to just let it bubble up, and catch it in the `ProgressMonitor` instead. Thoughts?

src/main/java/com/pinterest/secor/common/KafkaClient.java
```diff
@@ -212,6 +219,16 @@ public Message getCommittedMessage(TopicPartition topicPartition) throws Excepti
...
             return getMessage(topicPartition, committedOffset, consumer);
+        } catch (MessageDoesNotExistException e) {
+          // If a RuntimeEMessageDoesNotExistException exception is raised,
+          // the message at the last comitted offset does not exist in Kafka.
+          // This is usually due to the message being compacted away by the
+          // Kafka log compaction process.
+          //
+          // That is no an exceptional situation - in fact it can be normal if
+          // the topic being consumed by Secor has a low volume. So in that
+          // case, simply return null
```
HenryCaiHaiying: Needs to have Log.WARN in this case.

src/main/java/com/pinterest/secor/common/KafkaClient.java
HenryCaiHaiying: I am fine with either way.  The caller either needs to deal with that special exception or check null return value.

src/main/java/com/pinterest/secor/tools/ProgressMonitor.java
```diff
@@ -207,8 +207,14 @@ private void exportToStatsD(List<Stat> stats) {
                     assert committedOffset <= lastOffset: Long.toString(committedOffset) + " <= " +
                         lastOffset;
 
-                    long offsetLag = lastOffset - committedOffset;
-                    long timestampMillisLag = lastTimestampMillis - committedTimestampMillis;
+                    long offsetLag = 0L;
```
HenryCaiHaiying: Should the code just return around line 195?

If committed message is null, what will the value of lastMessage?  Will that also be null?  Otherwise secor would have another upload after the committed message.

src/main/java/com/pinterest/secor/common/KafkaClient.java
Fluxx: Ah yeah that's a good idea. Will do.

src/main/java/com/pinterest/secor/common/KafkaClient.java
Fluxx: Sure thing, will add.

src/main/java/com/pinterest/secor/tools/ProgressMonitor.java
Fluxx: Hmm...perhaps not return, but it could `continue` to skip this `TopicPartition`?

Fluxx:
Alright, changes have been made, including the `continue` one I hinted at. Please take another look.

HenryCaiHaiying:
Thanks for the contribution

PR #194, opened 2016-03-24
Microsoft Azure blob storage integration

tcnksm:
This PR  allows to use to Microsoft Azure blob storage service (https://azure.microsoft.com/en-us/documentation/articles/storage-introduction/#blob-storage) for backend.

PR #366, opened 2017-10-11
Document Microsoft Azure Blob Storage support

optiomas:
This commit documents the Microsoft Azure Blob Storage support
introduced in PR #194.


HenryCaiHaiying:
Looks good.  Thanks for the contribution.

PR #367, opened 2017-10-12
Use official Docker image "openjdk" instead of "java"

jethrocarr:
The official Java image is deprecated and has no longer recieved any updates since Dec 31 2016. As the Java image has always been built around OpenJDK, using the official Dockerhub OpenJDK image is a logical replacement.

See comments re deprecation at: https://hub.docker.com/r/library/java/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM java:8
+FROM openjdk:8
 
 RUN mkdir -p /opt/secor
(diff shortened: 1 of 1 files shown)


HenryCaiHaiying:
LGTM.  Thanks for the contribution.

jethrocarr:
My pleasure, thanks for all your work on Secor.

PR #371, opened 2017-10-19
JsonORCFileReaderWriterFactory crashes for null/non-existing values #369

attilagenc:
In my tests batch size was always 1 and my previous fix was to set index value to 0, correct fix should be to use "rowIndex" parameter as index.


HenryCaiHaiying:
LGTM

PR #376, opened 2017-11-13
Allow use of AWS session token

IvanAtanasov:
Small change to allow use of session token for S3 authentication.

This is required if you are using temporary access keys when running locally.

### DIFF
--- a/src/main/java/com/pinterest/secor/uploader/S3UploadManager.java
+++ b/src/main/java/com/pinterest/secor/uploader/S3UploadManager.java
@@ -16,6 +16,7 @@
  */
 package com.pinterest.secor.uploader;
 
+import com.amazonaws.auth.BasicSessionCredentials;
 import com.amazonaws.services.s3.model.ObjectMetadata;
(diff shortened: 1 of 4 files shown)


src/main/java/com/pinterest/secor/uploader/S3UploadManager.java
```diff
@@ -81,6 +82,7 @@ public S3UploadManager(SecorConfig config) {
 
         final String accessKey = mConfig.getAwsAccessKey();
         final String secretKey = mConfig.getAwsSecretKey();
+        final String sessionToken = mConfig.getAwsSessionToken();
```
HenryCaiHaiying: If the config is like 'aws.session.token=', will line 85 throw exception?

IvanAtanasov: It runs fine for me with it being set empty like that. Will just be an empty string similar to access key, secret key and role when they are not set

HenryCaiHaiying:
LGTM, thanks for the contribution

PR #382, opened 2017-12-04
Added a build profile for use with kafka version 0.10.0.1

IvanAtanasov:
the default of 0.10.2.0 is not working with our 0.10.0.1 kafka cluster. It produces java.nio.channels.ClosedChannelException exceptions whenever messages are added to be consumed

If there is a better solution than this, I am happy to use it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -598,7 +598,7 @@
                 <dependency>
                     <groupId>org.apache.kafka</groupId>
                     <artifactId>kafka_2.10</artifactId>
-                    <version>0.10.0.1</version>
+                    <version>0.10.2.0</version>
                     <exclusions>
                         <exclusion>
                             <groupId>org.slf4j</groupId>
@@ -640,6 +640,25 @@
         </profile>
         <profile>
             <id>kafka-0.10-dev</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+            </activation>
+            <dependencies>
+                <dependency>
+                    <groupId>org.apache.kafka</groupId>
+                    <artifactId>kafka_2.10</artifactId>
+                    <version>0.10.2.0</version>
+                    <exclusions>
+                        <exclusion>
+                            <groupId>org.slf4j</groupId>
+                            <artifactId>slf4j-simple</artifactId>
+                        </exclusion>
+                    </exclusions>
+                </dependency>
+            </dependencies>
+        </profile>
+        <profile>
+            <id>kafka-0.10.0.1</id>
             <activation>
                 <activeByDefault>true</activeByDefault>
             </activation>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -657,5 +657,24 @@
                 </dependency>
             </dependencies>
         </profile>
+        <profile>
+            <id>kafka-0.10.0.1</id>
```
HenryCaiHaiying: This version is actually lower than kafka-0.10.dev where it was using kafka 0.10.2.0, is there a particular reason that you need to use 0.10.0.1?

IvanAtanasov: I can't remember the specific errors but when trying to consume from a 0.10.0.1 cluster it does not work.

HenryCaiHaiying:
We probably prefer to keep less version profiles in the pom file, especially 0.10.0.1 is a deprecated version.

artem-garmash:
We have to do the same to make secor work with 0.10.0.1 clusters. The problem was in ZookeeperConnector. It was failing to fetch offsets and consuming would never start. The error message was:
```
Error while fetching offsets from 10.0.1.2:9092. Possible cause: null
```
While in kafka logs, there was the error about invalid version:
```
kafka.network.InvalidRequestException: Error getting request for apiKey: 9 and apiVersion: 2
        at kafka.network.RequestChannel$Request.liftedTree2$1(RequestChannel.scala:95)
        at kafka.network.RequestChannel$Request.<init>(RequestChannel.scala:87)
        at kafka.network.Processor$$anonfun$processCompletedReceives$1.apply(SocketServer.scala:488)
        at kafka.network.Processor$$anonfun$processCompletedReceives$1.apply(SocketServer.scala:483)
        at scala.collection.Iterator$class.foreach(Iterator.scala:893)
        at scala.collection.AbstractIterator.foreach(Iterator.scala:1336)
        at scala.collection.IterableLike$class.foreach(IterableLike.scala:72)
        at scala.collection.AbstractIterable.foreach(Iterable.scala:54)
        at kafka.network.Processor.processCompletedReceives(SocketServer.scala:483)
        at kafka.network.Processor.run(SocketServer.scala:413)
        at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.IllegalArgumentException: Invalid version for API key 9: 2
        at org.apache.kafka.common.protocol.ProtoUtils.schemaFor(ProtoUtils.java:31)
        at org.apache.kafka.common.protocol.ProtoUtils.requestSchema(ProtoUtils.java:44)
        at org.apache.kafka.common.protocol.ProtoUtils.parseRequest(ProtoUtils.java:60)
        at org.apache.kafka.common.requests.OffsetFetchRequest.parse(OffsetFetchRequest.java:117)
        at org.apache.kafka.common.requests.AbstractRequest.getRequest(AbstractRequest.java:52)
        at kafka.network.RequestChannel$Request.liftedTree2$1(RequestChannel.scala:92)
```
Better way to handle this would be to add kafka version property to the pom to allow easy version override when building secor.

HenryCaiHaiying:
OK, I will incorporate this change.

PR #387, opened 2018-01-10
Avro timestamp parsing and Avro to Parquet writing using Confluent Schema Registry, JSON to Parquet

achad4:
- Use Confluent’s CachedSchemaRegistryClient to allow Avro message parsing
- Add Avro to Parquet writing 
- Extend the current Protobuf reading and writing capabilities to allow writing Parquet files from JSON messages


src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -274,7 +276,7 @@ public String getAwsSecretKey() {
     }
 
     public String getAwsSessionToken() {
-        return getString("aws.session.token");
+        return getString("aws.session.token", "");
```
HenryCaiHaiying: Why this change?

src/main/java/com/pinterest/secor/common/SecorSchemaRegistryClient.java
```diff
@@ -0,0 +1,47 @@
...
+        } catch (Exception e){
+            LOG.error("Error initalizing schema registry", e);
+            throw new RuntimeException(e);
+        }
+    }
+
+    public GenericRecord decodeMessage(String topic, byte[] message) {
+        GenericRecord record = (GenericRecord) decoder.fromBytes(message);
+        Schema schema = record.getSchema();
+        schemas.putIfAbsent(topic, schema);
```
HenryCaiHaiying: Does this mean you will not update the topic schema to latest?

src/main/java/com/pinterest/secor/common/SecorConfig.java
achad4: I was unable to start the application when I didn't set this parameter. It shouldn't be required right?

src/main/java/com/pinterest/secor/common/SecorSchemaRegistryClient.java
achad4: That's a good point @HenryCaiHaiying, I will add unit tests to this class for that scenario.

achad4:
@HenryCaiHaiying thank you for reviewing! I managed to write JSON to parquet by extending the ProtobufUtil (see` Message decodeJsonMessage(String topic, byte[] payload)`). Per the documentation I added, this functionality will only kick in when explicitly set. 

Today, I actually extended some unit tests and made a couple changes which I plan to push in the morning. I'm looking forward to your feedback.

src/main/java/com/pinterest/secor/common/SecorSchemaRegistryClient.java
HenryCaiHaiying: So you don't want to update the schema?

src/main/java/com/pinterest/secor/common/SecorConfig.java
HenryCaiHaiying: We usually want those parameters to be explicitly set in your config file rather than java code defaulting.  But if you are not using aws, maybe we should not require this as a required config.

src/main/java/com/pinterest/secor/common/SecorSchemaRegistryClient.java
achad4: No, I do want to. I'll make changes and add tests to ensure they are updated. The SchemaRegistryClient should already handle changes, so this wrapper may even be unnecessary.

HenryCaiHaiying:
For the JSON->Parquet, it's a bit odd to use Protobuf, would JSON->Avro->Parquet work (that seems what most people was doing)?  In either situations, how do you specify the schema for JSON?

src/main/java/com/pinterest/secor/common/SecorConfig.java
achad4: I am using AWS but not a session token. Just permanent API keys and secrets. In that scenario I don't think think should be required.

achad4:
That's interesting, and I'd definitely be open to switching. I went with Protobuf, because Secor already had an established workflow from Protobuf to Parquet and I didn't want to add too much overhead.
You just define the protobuf schema as you normally would, and the JSON is mapped in memory to a protobuf message using this class https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/util/JsonFormat.Parser. 
The only additional configurations you need are to define which topics are using JSON (as described in the documentation I added). 
- `secor.topic.message.format.<topic>=JSON` or`secor.topic.message.format.*=JSON`

achad4:
@HenryCaiHaiying I'll push my JSON -> Protobuf changes and tests sometime tomorrow. I actually now have my latest version of this running in production on 6 JSON topics averaging together about 10000 messages per second.

src/main/java/com/pinterest/secor/common/SecorSchemaRegistryClient.java
achad4: @HenryCaiHaiying I fixed and tested this behavior: please see the new `SecorSchemaRegistryClientTest`

src/main/java/com/pinterest/secor/common/SecorConfig.java
HenryCaiHaiying: That's reasonable.

HenryCaiHaiying:
That's for the work, this is a good contribution.  Other people have waited for Avro and JSON for quite a while now.

achad4:
Thanks @HenryCaiHaiying! I don't think a pure avro implementation would be too tough.

PR #390, opened 2018-01-30
Bump libthrift to 0.10.0

thiagophx:
In order to use newly generated thrift classes we need to bump to a newer version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -255,7 +255,7 @@
         <dependency>
             <groupId>org.apache.thrift</groupId>
             <artifactId>libthrift</artifactId>
-            <version>0.5.0</version>
+            <version>0.10.0</version>
         </dependency>
         <dependency>
             <groupId>com.twitter</groupId>


HenryCaiHaiying:
This is a relative big version jump, are you able to run the unit/integration tests on your side?

thiagophx:
Hey, thanks for the quick answer.

The reason I created this PR is because newer versions of javabeans generated by thrift won't work with libthrift 0.5.0. It has different classes that the javabeans depend on.

The unittests run fine for me, but the integration fails (seems to be set up issue, not related to the change). I'm actually running this on my development env and it works fine, I can see parquet files generated from thrift on S3.

Is there any chance we could fix the build on travis? I believe it would be a better env to test it, because as I mentioned, I think I'm missing some proper set up for the integration tests.

Cheers!

HenryCaiHaiying:
Haven't got the time to fix the travis issue.  I think your thrift version change is probably fine, I checked our internal version was on 0.9.2, so 0.10 is probably fine.

PR #391, opened 2018-02-01
Fix ORC writer with ZLIB compression

artem-garmash:
The current version uses org.apache.orc.impl.ZlibCodec to enable ZLIB compression for ORC, but it's impossible to use it as secor fails to create it if it's specified in the secor.compression.codec property.
The PR uses org.apache.hadoop.io.compress.GzipCodec to trigger ORC ZLIB compression which is used usually for other formats.

### DIFF
--- a/src/main/java/com/pinterest/secor/io/impl/JsonORCFileReaderWriterFactory.java
+++ b/src/main/java/com/pinterest/secor/io/impl/JsonORCFileReaderWriterFactory.java
@@ -10,13 +10,13 @@ import org.apache.hadoop.hive.ql.exec.vector.VectorizedRowBatch;
 import org.apache.hadoop.io.compress.CompressionCodec;
 import org.apache.hadoop.io.compress.Lz4Codec;
 import org.apache.hadoop.io.compress.SnappyCodec;
+import org.apache.hadoop.io.compress.GzipCodec;
 import org.apache.orc.CompressionKind;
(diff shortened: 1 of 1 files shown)


HenryCaiHaiying:
I am not familiar with the difference between zlib and gzip.  It seems to me in most other big-data projects, GzipCodec is the commonly used one.  Is GzipCodec the superclass of ZliCodec?

With the change, do you specify gzip or zlib in your config file?

src/main/java/com/pinterest/secor/io/impl/JsonORCFileReaderWriterFactory.java
```diff
@@ -181,7 +181,7 @@ private CompressionKind resolveCompression(CompressionCodec codec) {
             return CompressionKind.LZ4;
         else if (codec instanceof SnappyCodec)
             return CompressionKind.SNAPPY;
-        else if (codec instanceof ZlibCodec)
+        else if (codec instanceof GzipCodec)
             return CompressionKind.ZLIB;
```
HenryCaiHaiying: Looks fine.  How about you add a comment here to explain why we would need to use GzipCodec to trigger ZLIB compression kind, then we are good to go.

artem-garmash:
ok, commented it

HenryCaiHaiying:
Thanks for the contribution.

PR #396, opened 2018-02-26
config: Clarify usage of secor.file.age.youngest

giner:


HenryCaiHaiying:
Looks good, thanks for the change.

PR #398, opened 2018-03-08
Don't swallow exception when trying to read bad config file

nenorbot:
When throwing an exception after failing to read the configuration file, don't ignore the underlying exception.


HenryCaiHaiying:

Issue #399, opened 2018-03-17
Filtering kafka messages


danpmx:
Hi, 
I have a requirement to filter out kafka messages before dumping them to storage
Is there any current support for that? I'm willing to contribute.
if not, where would be the best place to add this logic. 
I thought that maybe in consumer.java is there any other suggested

HenryCaiHaiying:
There is a Transformer interface you can transform a message before
persisting to S3.  It's called from Consumer.java

You can change the interface/contract a little bit by returning a null
message, a null message would mean filtering out in Consumer.java

PR #402, opened 2018-03-24
Add support for filtering messages out before dumping to storage

Danm87:
implementing #399


Danm87:
implementing #399

src/main/java/com/pinterest/secor/consumer/Consumer.java
```diff
@@ -139,9 +139,14 @@ protected boolean consumeNextMessage() {
             } catch (IOException e) {
                 throw new RuntimeException("Failed to adjust offset.", e);
             }
+
+            Message transformedMessage = mMessageTransformer.transform(rawMessage);
```
HenryCaiHaiying: Why the transform call is moved outside try/catch block?

Danm87:
by mistake, 
returned it back where it belonged :)

HenryCaiHaiying:
Thanks for the contribution.

PR #403, opened 2018-03-27
Schema registry value topic name

ryananguiano:
Add schema registry subject suffix and override


src/main/java/com/pinterest/secor/io/impl/AvroParquetFileReaderWriterFactory.java
```diff
@@ -108,9 +113,21 @@ public AvroParquetFileWriter(LogFilePath logFilePath, CompressionCodec codec) th
             CompressionCodecName codecName = CompressionCodecName
                     .fromCompressionCodec(codec != null ? codec.getClass() : null);
             topic = logFilePath.getTopic();
+
+            Schema schema;
```
HenryCaiHaiying: Does this logic also need to be replicated in AvroParquetFileReader?

ryananguiano: Done

ryananguiano:
Now getting an error when the file reader is checking upload policy.

```
2018-03-27 16:31:02,139 [Thread-5] (org.apache.parquet.hadoop.ParquetFileReader) INFO  Initiating action with parallelism: 5
2018-03-27 16:31:02,140 [Thread-5] (org.apache.parquet.hadoop.ParquetFileReader) INFO  reading another 1 footers
2018-03-27 16:31:02,141 [Thread-5] (org.apache.parquet.hadoop.ParquetFileReader) INFO  Initiating action with parallelism: 5
2018-03-27 16:31:02,191 [Thread-5] (org.apache.parquet.hadoop.InternalParquetRecordReader) INFO  RecordReader initialized will read a total of 26057 records.
2018-03-27 16:31:02,191 [Thread-5] (org.apache.parquet.hadoop.InternalParquetRecordReader) INFO  at row 0. reading next block
2018-03-27 16:31:02,258 [Thread-5] (org.apache.parquet.hadoop.InternalParquetRecordReader) INFO  block read in memory in 67 ms. row count = 26057
2018-03-27 16:31:02,353 [Thread-5] (com.pinterest.secor.main.ConsumerMain) ERROR Thread Thread[Thread-5,5,main] failed
java.lang.RuntimeException: Failed to apply upload policy
	at com.pinterest.secor.consumer.Consumer.checkUploadPolicy(Consumer.java:117)
	at com.pinterest.secor.consumer.Consumer.run(Consumer.java:107)
Caused by: org.apache.kafka.common.errors.SerializationException: Error deserializing Avro message for id -1
Caused by: org.apache.kafka.common.errors.SerializationException: Unknown magic byte!
```

It looks like https://github.com/revpoint/secor/blob/schema_registry_value_topic_name/src/main/java/com/pinterest/secor/uploader/Uploader.java#L168 is calling `AvroParquetFileReader` and is passing it back to `AvroParquetFileWriter`. This is calling:

`GenericRecord record = schemaRegistryClient.decodeMessage(topic, keyValue.getValue());`

I'm going to add a check to the writer, if there is no magic byte just pass the record as-is.

ryananguiano:
OK. `testAvroParquetReadWriteRoundTrip` now passes.

src/main/java/com/pinterest/secor/io/impl/AvroParquetFileReaderWriterFactory.java
```diff
@@ -66,6 +71,21 @@ public FileWriter BuildFileWriter(LogFilePath logFilePath, CompressionCodec code
         return serialized.array();
     }
 
+    protected Schema getSchema(String topic) {
+        Schema schema;
+        if (schemaSubjectOverride != null && !schemaSubjectOverride.isEmpty()) {
+            schema = schemaRegistryClient.getSchema(schemaSubjectOverride);
+        } else {
+            try {
+                schema = schemaRegistryClient.getSchema(topic);
```
HenryCaiHaiying: Is this call expensive?  Does it involve remote RPC?  this schemaRegistryClient.getSchema() was happening multiple times in this routine.

ryananguiano: getSchema() is only called on reader/writer creation. And this is done on purpose to allow for lookup of the original subject first, and if it doesn't find this then it will lookup the subject with "-value" appended, which is the default for certain schema registry clients like Python.

I believe this is the desired procedure for schema lookup.

ryananguiano: Also, it appears the project is using `CachedSchemaRegistryClient` so it should cache previous requests if it is created again.

ryananguiano: Or, if you want I can make the default null and then append the suffix only if it is set in config. This is more explicit, but will make it less likely to work "out of the box"

ryananguiano:
OK. Both settings are now explicit. I can also confirm working in production now.

src/main/java/com/pinterest/secor/io/impl/AvroParquetFileReaderWriterFactory.java
```diff
@@ -8,8 +8,8 @@
 import com.pinterest.secor.common.SecorSchemaRegistryClient;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericRecord;
-import org.apache.avro.io.Encoder;
-import org.apache.avro.io.EncoderFactory;
+import org.apache.avro.io.*;
```
HenryCaiHaiying: Don't use * in the import, please list each individual classes

src/main/java/com/pinterest/secor/io/impl/AvroParquetFileReaderWriterFactory.java
```diff
@@ -66,6 +73,28 @@ public FileWriter BuildFileWriter(LogFilePath logFilePath, CompressionCodec code
         return serialized.array();
     }
 
+    protected static GenericRecord deserializeAvroRecord(DecoderFactory decoderFactory, SpecificDatumReader<GenericRecord> reader, byte[] value) throws IOException {
```
HenryCaiHaiying: This line seems too long

src/main/java/com/pinterest/secor/util/AvroSchemaUtil.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.pinterest.secor.common.SecorConfig;
+
+public class AvroSchemaUtil {
+
+    public static String getAvroSubjectSuffix(SecorConfig config) {
+        return config.getString("avro.schema.subject.suffix", "");
+    }
+
+    public static String getAvroSubjectOverride(SecorConfig config) {
```
HenryCaiHaiying: Is this a global override or per-topic override?

src/main/java/com/pinterest/secor/io/impl/AvroParquetFileReaderWriterFactory.java
```diff
@@ -66,6 +73,28 @@ public FileWriter BuildFileWriter(LogFilePath logFilePath, CompressionCodec code
...
+        BinaryDecoder binaryDecoder = decoderFactory.binaryDecoder(value, null);
+        return reader.read(null, binaryDecoder);
+    }
+
+    protected GenericRecord decodeMessage(byte[] value, String topic, SpecificDatumReader<GenericRecord> reader) throws IOException {
+        if (value.length > 1 && value[0] == 0) {
```
HenryCaiHaiying: Where does this convention coming from (use the first byte to differentiate)?

src/main/java/com/pinterest/secor/util/AvroSchemaUtil.java
ryananguiano: This is a global override and has marked comments in the properties file:

```
# Setting this will override the subject lookup for all topics. This is only recommended if
# you are reading a single kafka topic.
avro.schema.subject.override=
```

src/main/java/com/pinterest/secor/io/impl/AvroParquetFileReaderWriterFactory.java
ryananguiano: I added a comment in the code:
Avro schema registry header format is a "Magic Byte" that equals 0 followed by a 4-byte int

src/main/java/com/pinterest/secor/io/impl/AvroParquetFileReaderWriterFactory.java
ryananguiano: Fixed

src/main/java/com/pinterest/secor/io/impl/AvroParquetFileReaderWriterFactory.java
ryananguiano: Fixed

ryananguiano:
Changes made

ryananguiano:
Is there anything else you need on this PR?

ryananguiano:
Topic override done

HenryCaiHaiying:
Is this PR in a finalized shape to be merged in?

ryananguiano:
This PR is ready to be merged. I am already running this branch in production

HenryCaiHaiying:
Can you close this PR and resubmit a new one?  I think the old PR still has JDK7 build inside, that's why it couldn't pass the travis integration tests.

PR #406, opened 2018-04-16
Added Simplaex to list of companies using secor

scravy:


HenryCaiHaiying:
Glad to see that the adopter of secor is increasing.

PR #408, opened 2018-04-20
Fixes usage of Gsuploadermanager

norrs:
Issues found while trying to use Gsuploadermanager as `secor.upload.manager.class` and running secor on kubernetes / GKE.


HenryCaiHaiying:
Thanks for the contribution.

PR #409, opened 2018-04-20
Enhancement in container runtime + Dockerfile issue with SSL certs

norrs:
Allow to set -Dconfig flag via env CONFIG_FILE on startup.  …
This allows us refer to another CONFIG_FILE during startup.  I've kept the
default which refer to `secor.prod.partition.properties` if not set.

In kubernetes we can simply control which config is used by volume mounting the
correct config map, but this keeps backwards compatibility and provides the
possibilities to:

 * All environments in one config map
 * A given environment in one config map.

Fixes missing cacerts bug currently happening in openjdk:8 image  …
This patch can be reverted once Debian Stable (Stretch) has backported the fix
mentioned in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=894979 .
See docker-library/openjdk#145 (comment)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 FROM openjdk:8
-
+# https://github.com/docker-library/openjdk/issues/145#issuecomment-334561903
+# https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=894979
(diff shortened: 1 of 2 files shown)


HenryCaiHaiying:
Thanks for the contribution.

Issue #394, opened 2018-02-08
Compile error on Java 7


gwilym:
Secor no longer compiles on Java 7 because of some use of Java 8+ features.

Full log: https://api.travis-ci.org/v3/job/338143233/log.txt

Sample:

```
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/travis/build/pinterest/secor/src/main/java/com/pinterest/secor/util/ProtobufUtil.java:[71,37] cannot find symbol
  symbol:   method putIfAbsent(java.lang.String,java.lang.String)
  location: variable messageFormatByTopic of type java.util.Map<java.lang.String,java.lang.String>
[ERROR] /home/travis/build/pinterest/secor/src/main/java/com/pinterest/secor/util/ProtobufUtil.java:[205,69] cannot find symbol
  symbol:   method getOrDefault(java.lang.String,java.lang.String)
  location: variable messageFormatByTopic of type java.util.Map<java.lang.String,java.lang.String>
[INFO] 2 errors 
```

(I think this was brought in with #387 @achad4)

Not sure what the direction for Secor is here. Should Secor be bumped to require Java 8, or should the code be updated to be compatible with Java 7?

HenryCaiHaiying:
I am inclined to deprecate the support for java7 since Oracle no longer supports it.

PR #410, opened 2018-04-20
Trying to fix integration tests

norrs:
Currently what I had to do locally on my development workstation to get integrations tests to run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ addons:
   hosts:
     - test-bucket.localhost
 env:
-  - PATH=$PATH:$HOME/.s3cmd SECOR_LOCAL_S3=true S3CMD=1.0.1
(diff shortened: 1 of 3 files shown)


norrs:
I would say this closes https://github.com/pinterest/secor/issues/394 as project contains JDK8 language features, and it seems you consider JDK7 support to be dropped.
 (I mean, secor is already deprecated in favor of merced that pinterest is using internally, which is planned to be open sourced soon™[1])

We've also been running our kafka servers with JDK8 without issues as well. 

[1] https://groups.google.com/forum/#!topic/secor-users/5kvLESSRBO0

HenryCaiHaiying:
Very nice change, thanks a lot for fixing this.  The integration tests was failing for quite a while.

PR #417, opened 2018-05-24
Provide more entrypoint options

### LINKED ISSUE #416: typo in echo command in docker-entrypoint.sh
This line here references the wrong bash variable (?)
https://github.com/pinterest/secor/blob/dcbc9b54c43e7ced03f09dccceeea47c56db94e6/src/main/scripts/docker-entrypoint.sh#L69

james-woods:
This PR allows more options to be defined through environment variables
when using the entrypoint script for Docker / K8s deployments.
It also fixes #416


HenryCaiHaiying:

PR #419, opened 2018-05-25
Kubernetes manifests and helm charts

norrs:
Hopefully this contribution can be useful for others wanting to try to run
things on kubernetes. Hopefully I should have managed to kept all in sync when removing zedge specific stuff. 

Hence this PR is 2 commits, since I don't really want to maintain the plain kubernetes manifests as helm is more "sane" after you scratch your head and get thru the horrible templating which gives flashback to 1994 and making of web pages ;-)


norrs:
Could be interesting to maybe put this into https://github.com/kubernetes/charts , but since secor is deprecated inside pinterest , not sure that is a thing to do..

norrs:
cc @stigsb

HenryCaiHaiying:
Thanks for the contribution.

valo:
@norrs thanks for providing helm charts for k8s. I am looking at deploying this on AWS right now and the charts will be helpful.

BTW, why is secor deprecated in pinterest?

james-woods:
I would love to know that as well - why has secor been deprecated?

norrs:
@valo @james-woods : I have done some more edits to the charts to fix issues with em, including updating some default values for kubernetes config files and I refactored the GCS Uploader Manager part to use the newest SDK and also had to add support for some throttling for one of our topics.

we also had to bump kafka versions etc and fix dependencies to match scala versions etc to get rid of some errors. Currently it is here, https://github.com/zedge/secor/tree/zedge-dev - gonna cherry pick and make new PR's into upstream SOON™ and sync the configuration values we have internally in our config repository. 

Reason for deprecation are mentioned by pinterest here: https://groups.google.com/forum/#!topic/secor-users/5kvLESSRBO0

cc @HenryCaiHaiying

norrs:
Note: we are in the verifying phase to ensure no logs get lost, I'll have an update next week.

valo:
Thanks for the reply. We decided to roll Secor on k8s without helm, but with using simple deployments for each topic job. So far it seems to be working fine. 

Seems like using a Kafka connect job is the more maintained path for such a service, although I am not sure why there is need to deal with yet another clustering solution when k8s can be used. Providing an easy way to run Secor in a k8s environment, might be a worthy extension of its functionality. I forked the project and made some adjustments... I need to think what is the best way to backport the changes into the main repo.

HenryCaiHaiying:
@valo Interested to see your work in this area.  Feel free to contribute back.

PR #420, opened 2018-06-01
Remove dotDash ./ so we can use fully qualified paths

norrs:
Avoid prefixing file with ./ so we can actually use fully qualified paths.
This is required in helm charts, which is gonna get an update soon™.


HenryCaiHaiying:
LGTM.

PR #421, opened 2018-06-02
Update source to work with kafka-1.0.0

norrs:
Also fixed CI to run with all profiles.

Changes:

You now need to specify a profile when building, because of scala versions clashes and kafka versions etc. 

I didn't get a default profile to work, as toggling it to another profile, the default one was still activated sadly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,8 +29,8 @@ Edit `src/main/config/*.properties` files to specify parameters describing the e
 
 ##### Create and install jars
 ```sh
-# By default this will install the "release" (Kafka 0.10 profile)
-mvn package
+# See pom.xml and select from valid profiles listed for your environment!
+mvn package -P <kafka_profile and matching scal version>
 mkdir ${SECOR_INSTALL_DIR} # directory to place Secor binaries in.
 tar -zxvf target/secor-0.1-SNAPSHOT-bin.tar.gz -C ${SECOR_INSTALL_DIR}
 
@@ -41,7 +41,7 @@ mvn -Pkafka-0.8-dev package
 ##### Run tests (optional)
 ```sh
 cd ${SECOR_INSTALL_DIR}
-./scripts/run_tests.sh
+MVN_PROFILE=<profile> ./scripts/run_tests.sh
 ```
 
 ##### Run Secor
--- a/README.md
+++ b/README.md
@@ -29,13 +29,13 @@ Edit `src/main/config/*.properties` files to specify parameters describing the e
 
 ##### Create and install jars
 ```sh
-# By default this will install the "kafka-0.11-1.0.0" (kafka 1.0.0, scala 2.11)
+# By default this will install the "kafka-0.10.2.0" profile
 mvn package
 mkdir ${SECOR_INSTALL_DIR} # directory to place Secor binaries in.
 tar -zxvf target/secor-0.1-SNAPSHOT-bin.tar.gz -C ${SECOR_INSTALL_DIR}
 
-# To use the Kafka 0.10.2.0 kafka libraries with scala 2.10 , use the release profile
-mvn -Prelease
+# To use the Kafka 1.0.0 kafka libraries with scala 2.11
+mvn -Pkafka-1.0.0
 ```
 
 ##### Run tests (optional)
--- a/README.md
+++ b/README.md
@@ -29,18 +29,20 @@ Edit `src/main/config/*.properties` files to specify parameters describing the e
 
 ##### Create and install jars
 ```sh
-# See pom.xml and select from valid profiles listed for your environment!
-mvn package -P <kafka_profile and matching scal version>
+# By default this will install the "kafka-0.11-1.0.0" (kafka 1.0.0, scala 2.11)
+mvn package
 mkdir ${SECOR_INSTALL_DIR} # directory to place Secor binaries in.
 tar -zxvf target/secor-0.1-SNAPSHOT-bin.tar.gz -C ${SECOR_INSTALL_DIR}
 
-# To use the Kafka 0.8 client you should use the kafka-0.8-dev profile
-mvn -Pkafka-0.8-dev package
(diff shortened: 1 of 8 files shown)


HenryCaiHaiying:
@norrs I was reading http://maven.apache.org/guides/introduction/introduction-to-profiles.html to see why the default profile is always included.  Based on the wiki:
...
All profiles that are active by default are automatically deactivated when a profile in the POM is activated on the command line or through its activation config.
...
I think secor's problem is there are two activeByDefault profiles: kafka-0.10.0.1 and kafka-0.10-dev, if you fix that, would that solved the problem?

It's OK to move the activeByDefault profile to 0.11.0 or 1.0.0, but I still want to see by default the developer don't have to specify profile for the default case.  There are other tricks by using environment variables to activate profile if we have to.

norrs:
@HenryCaiHaiying :

Default set to kafka-0.11-1.0.0 meaning scala 2_11 due to kafka version 1.0.0.

Fixed default profile and also added that travis just prints the mvn dependecy:tree in case we want to see that in build logs. 

```
$ MVN_PROFILE=release make dependency_tree | grep kafka
[INFO] +- io.confluent:kafka-avro-serializer:jar:2.0.1:compile
[INFO] |  +- io.confluent:kafka-schema-registry-client:jar:2.0.1:compile
[INFO] +- org.apache.kafka:kafka_2.10:jar:0.10.2.0:compile
[INFO] |  +- org.apache.kafka:kafka-clients:jar:0.10.2.0:compile

$ MVN_PROFILE=kafka-0.10.0.1 make dependency_tree | grep kafka
[INFO] +- io.confluent:kafka-avro-serializer:jar:2.0.1:compile
[INFO] |  +- io.confluent:kafka-schema-registry-client:jar:2.0.1:compile
[INFO] +- org.apache.kafka:kafka_2.10:jar:0.10.0.1:compile
[INFO] |  +- org.apache.kafka:kafka-clients:jar:0.10.0.1:compile

$ MVN_PROFILE=kafka-0.11-1.0.0 make dependency_tree | grep kafka
[INFO] +- io.confluent:kafka-avro-serializer:jar:2.0.1:compile
[INFO] |  +- io.confluent:kafka-schema-registry-client:jar:2.0.1:compile
[INFO] +- org.apache.kafka:kafka_2.11:jar:1.0.0:compile
[INFO] |  +- org.apache.kafka:kafka-clients:jar:1.0.0:compile

$ MVN_PROFILE=kafka-0.8-dev make dependency_tree | grep kafka
[INFO] +- io.confluent:kafka-avro-serializer:jar:2.0.1:compile
[INFO] |  +- io.confluent:kafka-schema-registry-client:jar:2.0.1:compile
[INFO] +- org.apache.kafka:kafka_2.10:jar:0.8.2.1:compile
[INFO] |  +- org.apache.kafka:kafka-clients:jar:0.8.2.1:compile


$ mvn dependency:tree  | grep kafka
[INFO] +- io.confluent:kafka-avro-serializer:jar:2.0.1:compile
[INFO] |  +- io.confluent:kafka-schema-registry-client:jar:2.0.1:compile
[INFO] +- org.apache.kafka:kafka_2.11:jar:1.0.0:compile
[INFO] |  +- org.apache.kafka:kafka-clients:jar:1.0.0:compile
```

norrs:
Note I kept the release profile to remain the same kafka 0.10.2.0 with scala 2.10 !

README.md
```diff
@@ -29,19 +29,21 @@ Edit `src/main/config/*.properties` files to specify parameters describing the e
 
 ##### Create and install jars
 ```sh
-# By default this will install the "release" (Kafka 0.10 profile)
+# By default this will install the "kafka-0.11-1.0.0" (kafka 1.0.0, scala 2.11)
```
HenryCaiHaiying: I thought you still make kafka-0.10.2 as the default?

norrs:
@HenryCaiHaiying : Regarding CI earlier, it simply was the of the builds who had a timeout when fetching maven dependency. So basically unlucky. 

Might reconsider if we need to test all the profiles and jdk versions, or maybe have 2 travis jobs, one for testing all profiles / jdk combinations  (which seems to now take about 2 hours :O ) and one that is only running with the release profile with openjdk.

HenryCaiHaiying:
For kafka-1.0.0 profile, do we still need to distinguish between scala_2.10 and scala_2.11?  For CI, the two travis profile options is interesting, but I don't know how much effort will it be to set it up?  I don't think we need to test different jdk versions, openjdk is fine.  So maybe three kafka profile + openjdk.  Also those profile runs are in parallel, so should more or less the same as one profile.

norrs:
@HenryCaiHaiying : I added it in the [README.dev.md](https://github.com/pinterest/secor/pull/421/files#diff-07f96ab9a374186187d89a4caba027b1), so that should be enough. Profiles should basically just be each kafka version we want to support when I gave it some more thought.  

Also we could maybe make some shell tests in another PR who just do some smart grepping in the `mvn dependency:tree` output. 

Regarding travis setup, it looks like it only supports having one `.travis.yml` file pr repo. This is the only thing they have about speeding up builds: https://docs.travis-ci.com/user/speeding-up-the-build/#Parallelizing-your-builds-across-virtual-machines 

And we can do something like this: https://github.com/travis-ci/travis-ci/issues/3540#issuecomment-330188464 for optimizing in another pr.

norrs:
@HenryCaiHaiying : I can update the `travis.yml` in this PR if you want to only do openjdk?

norrs:
@HenryCaiHaiying : If you want more changes, let me know.

HenryCaiHaiying:
Left a small comment on m2 cache.

norrs:
@HenryCaiHaiying : I Found it from the https://docs.travis-ci.com/user/caching/#Arbitrary-directories , but it actually seems to not be doing it's job?

norrs:
Removed the .m2 cache commit, as I didn't see the build time getting any faster ..

HenryCaiHaiying:
This is quite a significant change, thanks for the contribution

PR #422, opened 2018-06-02
GCP Cloud builder utilities and default yaml file

norrs:
For people who want to build secor with cloudbuilder

Sadly it currently builds as root *screams*.


HenryCaiHaiying:
Don't have GCS environment to verify, but will take it base on faith.

PR #423, opened 2018-06-02
Gcs uploadermanager refactor

norrs:
Should be reviewed after https://github.com/pinterest/secor/pull/421 as it's rebased on top of this PR.

Also would be nice to get in touch with other GCS users to verify it works for em. This seems to be working for us, but had to add support for throttling towards GCS API usage due to:

> If you get exceptions towards uploading with:

>com.google.cloud.storage.StorageException: Error writing request body to
>server
>
>This is most likley due to you getting throttled / their API respons
>429/5xx. (?) Exception sadly doesn't say anything which is sad library. 
>
>See https://cloud.google.com/storage/docs/request-rate as well.

Throttling it , helped us at least. Most likley due to the time Google need to scale the bucket for more QPS if needed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -338,7 +338,7 @@
         <dependency>
             <groupId>com.google.cloud</groupId>
             <artifactId>google-cloud-storage</artifactId>
-            <version>1.31.0</version>
+            <version>1.33.0</version>
         </dependency>
         <dependency>
             <groupId>com.microsoft.azure</groupId>
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -480,6 +480,10 @@ public class SecorConfig {
         return getString("secor.gs.path");
     }
 
+    public double getGsRateLimit() {
+        return getDouble("secor.gs.tasks.ratelimit.pr.second", 10.0);
+    }
+
     public int getGsConnectTimeoutInMs() {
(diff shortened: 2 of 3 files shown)


HenryCaiHaiying:
Looks to me that the code to calling GS upload changed quite a bit, are they equivalent?  I don't have GS environment to try it out.

norrs:
@HenryCaiHaiying : This is the implementation we are using and works for us, and this uses the latest cloud storage client sdk's provided by google. 

https://github.com/GoogleCloudPlatform/google-cloud-java and https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-clients/google-cloud-storage

You maybe could have kept both implementations, but this requires more matrixes with profiles I assume to avoid dependency clashes.

HenryCaiHaiying:
OK, I will wait for #421 being merged first.

norrs:
Rebased on top of latest changes in #421 , commit https://github.com/pinterest/secor/pull/421/commits/7bd0207bc2064ba60647aaad75aefaf9d4660974

norrs:
Rebased again, removing the non-working m2 cache stuff in .travis file.

norrs:
Hm, so even tho you merged #421 I had to rebase on top of the merge commit again in master. So here we go again @HenryCaiHaiying . Sorry about the notifications. Sighs.

HenryCaiHaiying:
It should be safe to merge now, thanks for the patience.

PR #424, opened 2018-06-03
support milliseconds long for TimestampColumnVector

lukess:
Add support long type milliseconds JSON number field in VectorColumnFiller for ORC timestamp data type.

### DIFF
--- a/src/main/java/com/pinterest/secor/util/orc/VectorColumnFiller.java
+++ b/src/main/java/com/pinterest/secor/util/orc/VectorColumnFiller.java
@@ -123,7 +123,7 @@ public class VectorColumnFiller {
                             new Timestamp(value.getAsLong()));
                 } else {
                     if (!back.isBackOff()) {
-                        LOG.warn("Timestamp format is neither String nor Number");
+                        LOG.warn(String.format("Timestamp format is neither String nor Number: %s", value.toString()));
                     }
                     vect.noNulls = false;
                     vect.isNull[row] = true;
--- a/src/main/java/com/pinterest/secor/util/orc/VectorColumnFiller.java
+++ b/src/main/java/com/pinterest/secor/util/orc/VectorColumnFiller.java
@@ -123,7 +123,7 @@ public class VectorColumnFiller {
                             new Timestamp(value.getAsLong()));
                 } else {
                     if (!back.isBackOff()) {
-                        LOG.warn(String.format("Timestamp format is neither String nor Number: %s", value.toString()));
+                        LOG.warn("Timestamp is neither string nor number: {}", value);
                     }
                     vect.noNulls = false;
                     vect.isNull[row] = true;
--- a/src/main/java/com/pinterest/secor/util/orc/VectorColumnFiller.java
+++ b/src/main/java/com/pinterest/secor/util/orc/VectorColumnFiller.java
@@ -3,6 +3,7 @@ package com.pinterest.secor.util.orc;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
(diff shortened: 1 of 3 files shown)


src/main/java/com/pinterest/secor/util/orc/VectorColumnFiller.java
```diff
@@ -105,11 +104,21 @@ public void convert(JsonElement value, ColumnVector vect, int row) {
...
+                            row,
+                            Timestamp.valueOf(value.getAsString().replaceAll(
+                                    "[TZ]", " ")));
+                } else if (value.getAsJsonPrimitive().isNumber()) {
+                    TimestampColumnVector vector = (TimestampColumnVector) vect;
+                    vector.set(
+                            row,
+                            new Timestamp(value.getAsLong()));
+                } else {
+                    vect.noNulls = false;
```
HenryCaiHaiying: Do you want to issue some warnings in this case?  If you worried about too much logging being emitted, you can so some exponential back-off logging (e.g. logging on 1th, 2nd, 4th, 8th ... occurrences), those are very useful to debugging.

lukess: @HenryCaiHaiying thanks for response. Yeah I did check the logging from VectorColumnFiller.java as well, but couldn't find any in that package path. I do not worry about too much logging because log would be only debug level (for only debugging as well). I'm going to try back-off logging as well 👍

HenryCaiHaiying: I think the problem is you don't any loggings when it gets to 'else' part of the code, usually this kind of logging is probably on warning level.  And to avoid generate too much warnings in the log file, you can try to log with exponential backoff.

src/main/java/com/pinterest/secor/util/orc/VectorColumnFiller.java
```diff
@@ -100,16 +103,31 @@ public void convert(JsonElement value, ColumnVector vect, int row) {
...
+                            Timestamp.valueOf(value.getAsString().replaceAll(
+                                    "[TZ]", " ")));
+                } else if (value.getAsJsonPrimitive().isNumber()) {
+                    TimestampColumnVector vector = (TimestampColumnVector) vect;
+                    vector.set(
+                            row,
+                            new Timestamp(value.getAsLong()));
+                } else {
+                    if (!back.isBackOff()) {
+                        LOG.warn("Timestamp format is neither String nor Number");
```
HenryCaiHaiying: Do you want to show the error 'value' value (and with its type as well)?  It will help the debugging.

HenryCaiHaiying:
Thanks for the introducing that Backoff class, looks cool.

src/main/java/com/pinterest/secor/util/orc/VectorColumnFiller.java
```diff
@@ -123,7 +123,7 @@ public void convert(JsonElement value, ColumnVector vect, int row) {
                             new Timestamp(value.getAsLong()));
                 } else {
                     if (!back.isBackOff()) {
-                        LOG.warn("Timestamp format is neither String nor Number");
+                        LOG.warn(String.format("Timestamp format is neither String nor Number: %s", value.toString()));
```
HenryCaiHaiying: I thought usually you will log using this format:

LOG.warn("Timestamp is neither string nor number: {}", value);

lukess: @HenryCaiHaiying you are correct. I should follow that.

HenryCaiHaiying:
LGTM

HenryCaiHaiying:
Thanks for the contribution.

PR #426, opened 2018-06-06
Honor Message Timestamp Name Separator Env Variable (#425)

### LINKED ISSUE #425: Message Timestamp Name Separator env variable is ignored in Dockerfile
Due to a typo (?) in the current Dockerfile the env variable `MESSAGE_TIMESTAMP_SEPARATOR` has no effect - it should address `message.timestamp.name.separator` not `message.timestamp.separator`

see:
https://github.com/pinterest/secor/blob/dae0fc2639e525568e542564986214a227741d5b/src/main/scripts/docker-entrypoint.sh#L76

james-woods:
This PR corrects the env variable that is configuring the timestamp separator in JSON messages. It fixes #425


HenryCaiHaiying:
Thanks for the contribution.

PR #427, opened 2018-06-08
Changes for making the secor work with ap-south-1 region of aws

rajagopalvreghunath:
Changes for making the secor work with ap-south-1 region of aws

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -126,12 +126,12 @@
         <dependency>
             <groupId>com.amazonaws</groupId>
             <artifactId>aws-java-sdk-s3</artifactId>
-            <version>1.11.313</version>
+	     <version>1.11.342</version>
         </dependency>
         <dependency>
             <groupId>com.amazonaws</groupId>
             <artifactId>aws-java-sdk-sts</artifactId>
-            <version>1.11.313</version>
+            <version>1.11.342</version>
         </dependency>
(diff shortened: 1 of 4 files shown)


HenryCaiHaiying:
Thanks for the contribution, but it looks like CI test didn't pass, did you look into that?

rajagopalvreghunath:
cool i ll check

norrs:
@rajagopalvreghunath :

```
ERROR Log file verifier failed
java.io.IOException: The s3n:// client to Amazon S3 is no longer available: please migrate to the s3a:// client
```
https://wiki.apache.org/hadoop/AmazonS3: 
> important: S3A requires the exact version of the amazon-aws-sdk against which Hadoop was built (and is bundled with). If you try to upgrade the library by dropping in a later version, things will break. 

You need to check  and verify what aws-java-sdk-s3 is pulling in of hadoop dependencies, as they have to be matching. 

Note that tests are using https://github.com/jubos/fake-s3 in travis to do it's integration checks, but don't thing that's where the error is. Somehow the new libraries tries to use `s3n://` protocol instead of the `s3a://` protocol..

HenryCaiHaiying:
Closing for now, @rajagopalvreghunath if you can figure out the test failures, feel free to re-open it.

PR #428, opened 2018-06-08
Add avro writing capabilities

achad4:
-- Due to the high memory requirements of transforming Avro - Parquet, I added the ability to write out Avro files directly

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -430,18 +430,6 @@
                             </descriptors>
                         </configuration>
                     </execution>
-                    <execution>
-                        <id>create-jar-with-deps</id>
-                        <phase>package</phase>
-                        <goals>
-                            <goal>single</goal>
-                        </goals>
-                        <configuration>
-                            <descriptorRefs>
-                                <descriptorRef>jar-with-dependencies</descriptorRef>
-                            </descriptorRefs>
-                        </configuration>
-                    </execution>
                 </executions>
 		    </plugin>
 			<!-- Protobuf source generator plug-in -->
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,14 @@
             <id>confluent</id>
             <url>http://packages.confluent.io/maven/</url>
         </repository>
+        <repository>
+            <id>central</id>
+            <name>Maven Repository Switchboard</name>
+            <layout>default</layout>
+            <url>http://repo1.maven.org/maven2/</url>
+        </repository>
     </repositories>
+
     <pluginRepositories>
         <pluginRepository>
             <id>Twitter public Maven repo</id>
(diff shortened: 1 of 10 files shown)


pom.xml
```diff
@@ -430,6 +430,18 @@
                             </descriptors>
                         </configuration>
                     </execution>
+                    <execution>
+                        <id>create-jar-with-deps</id>
```
HenryCaiHaiying: That does this create-jar-with-deps do?  Will it make one big gigantic jar?

achad4: I can remove it (I use it for our setup at Sailthru) if you want. It creates an uber jar with all the dependencies

norrs: If you want to make a fatjar, isn't using maven-shadow-plugin the way to go? 

something like:

```
         <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-shade-plugin</artifactId>
                <configuration>
                    <!-- DO NOT MINIMIZE JARS -->
                    <minimizeJar>false</minimizeJar>
                    <shadedArtifactAttached>false</shadedArtifactAttached>
                    <createDependencyReducedPom>false</createDependencyReducedPom>
                    <keepDependenciesWithProvidedScope>false</keepDependenciesWithProvidedScope>
                    <useBaseVersion>true</useBaseVersion>
                    <filters>
                        <filter>
                            <artifact>*:*</artifact>
                            <excludes>
                                <!--
                                  I have no idea why adding HTML to main jar files have
                                  any benefit. But it seems pretty common.
                                 -->
                                <exclude>**/package.html</exclude>
                                <exclude>**/overview.html</exclude>
                                <exclude>about.html</exclude>
                                <!--
                                  The generated package-info.class is usually *not* included
                                  in compiled jars, but some projects have it in their jar file.
                                  It is never needed, and cannot be loaded as a class anyway.
                                 -->
                                <exclude>**/package-info.class</exclude>
                            </excludes>
                        </filter>
                    </filters>
                </configuration>
                <executions>
                    <!-- Run shade goal on package phase -->
                    <execution>
                        <phase>package</phase>
                        <goals>
                            <goal>shade</goal>
                        </goals>
                        <configuration>
                            <finalName>${project.artifactId}</finalName>
                            <transformers>
                                <!-- add Main-Class to manifest file -->
                                <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
                                    <mainClass>${main.class}</mainClass>
                                </transformer>
                            </transformers>
                        </configuration>
                    </execution>
                </executions>
            </plugin>

```

https://maven.apache.org/plugins/maven-shade-plugin/ & https://maven.apache.org/plugins/maven-shade-plugin/usage.html

This code above is just copied and slightly modified from what we've used internally, so need to be adjusted to be working with this project.

HenryCaiHaiying: I prefer not having a fatjar by default.

achad4:
Hey @HenryCaiHaiying I removed the "create-jar-with-deps" step and fixed the failing tests

HenryCaiHaiying:

PR #429, opened 2018-06-09
parse nested timestamps

rongo89:
enable possibility to parse nested timestamps in avro messages

### DIFF
--- a/src/main/config/secor.ak.part.properties
+++ b/src/main/config/secor.ak.part.properties
@@ -1,0 +1,385 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


src/main/java/com/pinterest/secor/parser/AvroMessageParser.java
```diff
@@ -31,7 +32,23 @@ public long extractTimestampMillis(final Message message) {
...
+                String timestampName = mConfig.getMessageTimestampName();
+                Object fieldValue;
+                
+                if (delimiter != null && !delimiter.isEmpty()) {
+                  String separatorPattern = Pattern.quote(delimiter);
+                  String[] fieldArray = timestampName.split(separatorPattern);
+                  int nameArrayLength = fieldArray.length - 1;
+
+                  for(int i = 0; i < nameArrayLength; i++) {
+                    record = (GenericRecord) record.get(fieldArray[i]);
```
HenryCaiHaiying: Do you want to generate some log warnings if some of the fields are not present in the record?

rongo89: Might make sense in case timestampRequired is false. I will add it.

pom.xml
```diff
@@ -6,7 +6,7 @@
 
     <groupId>com.pinterest</groupId>
     <artifactId>secor</artifactId>
-    <version>0.26-SNAPSHOT</version>
+    <version>0.26</version>
```
HenryCaiHaiying: The current value is: 0.27-SNAPSHOT

src/main/config/secor.ak.part.properties
```diff
@@ -0,0 +1,385 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
HenryCaiHaiying: How is this secor.ak.part.properties being used?

HenryCaiHaiying:
I saw this is closed, were you planning to merge this PR to the main secor repo?

PR #432, opened 2018-06-11
Updates to helm charts

norrs:
New log4j.kubernetes-prod.properties config:
Don't want to have "Im using this compression" for every upload.

Added the refactored GCS upload manager configs for avoid flooding GCS
API.

Max age to 400 seconds, what we currently use in production and these
config values works with on kubernetes for us.

Offsets in zookeeper instead of kafka, due to other main programs which
reads offsets from kafka for keeping track.

In helm charts:

Support for adding resources, ie requests and limits from values in
helm.

Specific log4j config files.

JVM memory

and yes, requests lower then jvm volume looks weird, but we allow to
overcommit on some memory usage to utilize nodes better in kubernetes.
Still under review and testing. Currently still observed lower memory
then what is requested..

Important config value to know about: secor.messages.per.second
Increasing this gave us way higher oommmffff consuming from kafka :-)

### DIFF
--- a/deploys/helm/secor/config/secor.kubernetes-prod.properties
+++ b/deploys/helm/secor/config/secor.kubernetes-prod.properties
@@ -115,6 +115,18 @@ swift.api.key=
 # https://cloud.google.com/storage/docs/json_api/v1/how-tos/upload
 secor.gs.upload.direct=false
 
+# Limit the amonut of API calls towards GCS
+# This usually helps against exceptinos of: com.google.cloud.storage.StorageException: Error writing request body to server
+# Cold/Low traffic buckets need to be "warmed" up most likely before you can increase this value.. this is untested.
+# Sadly the client library doesn't say which http response code happens, ie 429 / 5xx etc.
+# Also see https://cloud.google.com/storage/docs/request-rate
+secor.gs.tasks.ratelimit.pr.second=2
+
+# Old behavior of number of threads was 256 , but running this on kubernetes we experience we wanted less threads here,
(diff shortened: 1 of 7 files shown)


deploys/helm/secor/config/secor.kubernetes-prod.properties
```diff
@@ -181,7 +193,7 @@ kafka.dual.commit.enabled=false
 
 # Storage offset.
 # Possible values: "zookeeper" to read offset from zookeeper or "kafka" to read offset from kafka consumer topic
-kafka.offsets.storage=kafka
+kafka.offsets.storage=zookeeper
```
HenryCaiHaiying: Is the storage switch just for your use case or do you think by default all k8 installations should use ZK storage?

norrs: @HenryCaiHaiying :

In newer kafka libraries with the new consumer code, it by default stores offsets in kafka itself I think.

Default before in this project has been what's defined in the `src/main/config/secor.common.properties` which is `zookeeper`.

I did try it with `kafka` a bit, but returned back to using `zookeeper` as this is what the `com.pinterest.secor.main.ProgressMonitorMain` uses for monitoring offsets.. 

So until we write to new consumer library code, I think it might be better to stick with `zookeeper` or hear what others have been testing this with.

So this is my reasining for at least running it with zk storage ourselves.

HenryCaiHaiying:
Makes sense, secor was using ZK storage all along, we haven't tested with kafka storage yet.

PR #435, opened 2018-06-20
commit offset to Kafka when offsets.storage = kafka

lukess:
Commit offset to Kafka when offsets.storage is kafka which is __consumer_offsets topic.

### DIFF
--- a/src/main/java/com/pinterest/secor/uploader/Uploader.java
+++ b/src/main/java/com/pinterest/secor/uploader/Uploader.java
@@ -21,6 +21,7 @@ import com.pinterest.secor.common.FileRegistry;
 import com.pinterest.secor.common.LogFilePath;
 import com.pinterest.secor.common.OffsetTracker;
 import com.pinterest.secor.common.SecorConfig;
+import com.pinterest.secor.common.SecorConstants;
 import com.pinterest.secor.common.TopicPartition;
 import com.pinterest.secor.common.ZookeeperConnector;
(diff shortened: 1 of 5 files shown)


src/main/java/com/pinterest/secor/uploader/Uploader.java
```diff
@@ -113,7 +122,9 @@ protected void uploadFiles(TopicPartition topicPartition) throws Exception {
                 mFileRegistry.deleteTopicPartition(topicPartition);
                 mZookeeperConnector.setCommittedOffsetCount(topicPartition, lastSeenOffset + 1);
                 mOffsetTracker.setCommittedOffsetCount(topicPartition, lastSeenOffset + 1);
-
+                if (mConfig.getOffsetsStorage().equals("kafka")) {
```
HenryCaiHaiying: Can this constant "kafka" defined in a a central place, e.g. SecorConfig.java?  And should it be 'equals' or 'equalsIgnoreCase'?

lukess: @HenryCaiHaiying yeah I was actually thinking the same thing. I think Constants way is better.

HenryCaiHaiying: Also make this check in uploader.init method, this check should only be done once since the config doesn't change.

HenryCaiHaiying:
Looks good to me, thanks for the contribution

Issue #397, opened 2018-02-28
Deprecated ConsumerConnector


efeller:
Looks like Secor is using a deprecated Kafka consumer API. As a result of this I am unable to use SASL options like "sasl.mechanism". VerifiableProperties reports:

2018-02-28 00:27:15,575 [Thread-6] (kafka.utils.VerifiableProperties) WARN  Property sasl.mechanism is not valid
2018-02-28 00:27:15,575 [Thread-5] (kafka.utils.VerifiableProperties) WARN  Property sasl.jaas.config is not valid
2018-02-28 00:27:15,576 [Thread-6] (kafka.utils.VerifiableProperties) WARN  Property security.protocol is not valid

Would be great if we could have a Secor version using the new consumer API and/or have some alternative solution/workaround for this. Thanks.

orderedchaosdev:
This is the same as https://github.com/pinterest/secor/issues/393

HenryCaiHaiying:
If guys have time, you can contribute a change to use the new consumer API,
I don't think it would be particularly difficult.

efeller:
@HenryCaiHaiying @orderedchaosdev Thanks for the quick response. We have a fork of Secor here. I will try to find some time to write a patch and hopefully contribute it back.

HenryCaiHaiying:
Thanks a lot, let me know when you are ready to contribute.

chitaks:
@efeller @HenryCaiHaiying Any updates?

igorcalabria:
@chitaks @tkaymak I'll submit a pull request this week. I'll tag you guys when its up

Issue #393, opened 2018-02-07
Kafka authentication support?


aig787:
I looked through the sample configuration and the code and didn’t see any way to use TLS authentication when connecting to the Kafka brokers. Is there any way to do this?

gwilym:
Secor uses Kafka's SimpleConsumer API which, as far as I can tell, has no support for SSL/TLS settings, and doesn't look like it will be updated to support them according to: https://issues.apache.org/jira/browse/KAFKA-2440

This is something I've been looking into, but I can't make any timeframe guarantees, so if anyone else can tackle it feel free to jump in.

HenryCaiHaiying:
for secor, the main data flow is going through high level consumer.  We
used simple consumer for fetch a particular message for monitoring and
partition finalization check.

TLS support is added in kafka 0.9, it should be doable to add TLS
support/configuration for secor which already supports 0.10.  Just need
someone to contribute a PR.

gwilym:
I'm looking at adding support for loading an additional .properties file for kafka consumer configuration, so that it's not necessary to repeat all the kafka properties in secor's configs. Properties from that additional file would be merged in to and take precedence over anything inferred from secor's standard configs. That should help with situations like this, and with forward compatibility.

I was blocked by #394 though since I need to use jitpack.io to use my fork as a dependency in a docker build (and it tries and fails to build java 7, since it infers requirements from .travis.yml).

gwilym:
I can drop java 7 support in my branch at least to unblock myself.

HenryCaiHaiying:
How about you remove the openjdk7 in .travis.yml to unblock you?

For the properties file, how many properties are there?

gwilym:
I'm not sure I understand the heart of the question, as the number is not predictable. It would depend on how the user needs to tune the Kafka consumer to their requirements.

In the specific case of enabling SSL support, you might for example need:

```
security.protocol=?
ssl.enabled.protocols=?
ssl.key.password=?
ssl.keystore.location=?
ssl.keystore.password=?
ssl.truststore.location=?
ssl.truststore.password=?
```

These are options I certainly need to add (at a minimum) for my own SSL with client-certificate authentication needs.

HenryCaiHaiying:
That still seems manageable within the current secor config file.  the
kafka parameters usually takes the form of 'kafka.min.fetch.bytes' etc.  If
you want to generalize, you can say putting every config prefixed with
'kafka.' into ConsumerConfig, but that that break something.

[...]

Issue #447, opened 2018-08-03
Using offset storage in Kafka and not Zookeeper not working


ronhash10:
Hey,
We are trying to move from Zookeeper to Kafka offset storage:
```
kafka.offsets.storage=kafka
kafka.dual.commit.enabled=false
```
We see though, that two consumers are created, one in Zookeeper, and one in Kafka, but the one that secor uses to commit and read the offset from is the Zookeeper.
Is there any support for having offset storage in Kafka in Secor?
Thanks,
Ron.

HenryCaiHaiying:
you need to set two params, take a look at: https://github.com/pinterest/secor/commit/c3c9b2dc745b6be858e070ff5d8d04f768d11ba9#diff-5bb7557074535d902a0519f1c5faeb23

HenryCaiHaiying:
I feel secor will always use zookeeper unless the code in ZookeeperConnector changes.  You can probably use dual commit to commit the offsets to both __consumer_offsets topic and ZK

sjoerdmulder:
We are facing the same issue; it seems that the kafka offset storage is not working and not used; It's only creating a zookeeper one. Which could also be seen here: https://github.com/pinterest/secor/blob/5d2fc90d6725f6abf8cb0acdbfc9f23187d3a131/src/main/java/com/pinterest/secor/uploader/Uploader.java#L114

HenryCaiHaiying:
I don't think secor is going to read from kafka offset storage unless these
code be changed.

But on the other hand, even if you moved offset storage to kafka, the ZK
dependency will still be there (kafka brokers will still need ZK to store
lots of other metadata information).

juhoautio:
I also tried migrating from zookeeper to kafka offsets storage, but got data loss. How can I migrate the offset storage without losing any data?

Secor seems to commit offsets to both kafka & zookeeper now (all partitions), but file uploading seems to be off sync. It could be related to this condition:

https://github.com/pinterest/secor/blob/c457ed4df255d5dcc9ce78ab6bd07f7faec3d08b/src/main/java/com/pinterest/secor/uploader/Uploader.java#L113-L116

----

Previous conf:

```properties
kafka.dual.commit.enabled=false
kafka.offsets.storage=zookeeper
```

Changed like this and restarted secor:

```properties
kafka.dual.commit.enabled=true
kafka.offsets.storage=kafka
```

[...]

PR #450, opened 2018-08-30
Kafka 2.0.0 Client

igorcalabria:
https://github.com/pinterest/secor/issues/397 https://github.com/pinterest/secor/issues/393
I wasn't sure If maintaining compatibility between all kafka versions was desired but I figured that we can't simply change a bunch of stuff without giving people some time to adjust. Release profile is still set to 0.10.2. 

It is important to highlight some things:
- New kafka client lib does not include the broker, so we need another way to run it. I used `docker-compose` because it was easier for me and it made some sense because the project already supports docker. This is not set in stone and we can change that

- I don't think we'll be able to support committing to kafka with the new client. After a rebalance, kafka will throw `CommitFailedException` if you try to commit a rebalanced TopicPartition. These topic partitions will be assigned to another instance but `commit()` will never be called in those instances since secor will skip already uploaded files(locked via zookeeper). Because of this, I'm simply reading the offsets from zookeeper.

- Travis will fail because I haven't configured docker-compose for it. We need to agree on a way to run kafka in the integration tests

- `KafkaMessageIterator` interface fits pretty weirdly with the new consumer's `poll` method. I Chose this approach because it minimizes the changes I had to do to the old code. 

- I haven't added all new consumer configs to `SecorConfig`. "low importance" ones were skipped https://kafka.apache.org/documentation/#configuration

- Performance test and production validation. I'll deploy a production copy in our environment and see how the new consumer behaves. I'll try to force rebalances as much as possible. It would be great if more people ran something similar.


src/main/java/com/pinterest/secor/reader/SecorKafkaMessageIterator.java
```diff
@@ -0,0 +1,145 @@
...
+                Map<TopicPartition, Long> committedOffsets = getCommittedOffsets(collection);
+                committedOffsets.forEach(((topicPartition, offset) -> {
+                    if (offset == -1) {
+                        if (offsetResetConfig.equals("earliest")) {
+                            mKafkaConsumer.seekToBeginning(collection);
+                        } else if (offsetResetConfig.equals("latest")) {
+                            mKafkaConsumer.seekToEnd(collection);
+                        }
+                    } else {
+                        LOG.debug("Seeking {} to offset {}", topicPartition, Math.max(0, offset));
```
HenryCaiHaiying: Is this a kafka bug?  I would imagine the kafka consumer will reset to the correct offset after partition reassignment.  And also do we need to clear any data structure and local data file because of the offset change?

src/main/java/com/pinterest/secor/tools/TestLogMessageProducer.java
```diff
@@ -55,17 +56,15 @@ public TestLogMessageProducer(String topic, int numMessages, String type,
     public void run() {
         Properties properties = new Properties();
         if (mMetadataBrokerList == null || mMetadataBrokerList.isEmpty()) {
-            properties.put("metadata.broker.list", "localhost:9092");
+            properties.put("bootstrap.servers", "localhost:9092");
```
HenryCaiHaiying: Will this property name change break the tests under older kafka version?

src/main/java/com/pinterest/secor/reader/SecorKafkaMessageIterator.java
igorcalabria: We need this because new kafka consumers don't support storing offsets in zookeeper. After an assignment, we need to manually seek to the offsets stored there[0]. The decision to keep using zookeeper came from two major points:
- People will be able to upgrade to a new secor version without having to migrate their commited offsets to kafka storage
- New kafka consumers commit model doesn't fit well with secor, in a nutshell, `consumer.commit` can throw a non retryable exception called `CommitFailedException`[1]. If this happens, we can either ignore it and end up with duplicate data(partitions were assigned to another consumer) or we can try to undo our action(deleting the file from s3). I thought that implementing this functionality is risky and not worth it since we can just keep using the "old" model of commiting to zookeeper.

[0] - https://kafka.apache.org/20/javadoc/org/apache/kafka/clients/consumer/ConsumerRebalanceListener.html
[1] - https://kafka.apache.org/20/javadoc/org/apache/kafka/clients/consumer/CommitFailedException.html

src/main/java/com/pinterest/secor/tools/TestLogMessageProducer.java
igorcalabria: It doesn't break the tests because I migrated this class to use the new producer. I was able to do this migration because the old consumer dependency also includes the new consumer

igorcalabria:
About cleanup on reassignment:
We don't actually need to do anything there because the old consumers behaved the same way and secor already deals with it(trimming files is an example of such checks)

About integration tests:
I run tests for both the legacy consumer and the new one. I'll update this PR with minor fixes and a working `.travis.yml`. Do you think its ok to add docker-compose? Its going to be pretty weird right now because of all the packaging and moving stuff to /tmp folders but eventually those integration tests could be refactored to be based around docker-compose.

src/main/java/com/pinterest/secor/reader/SecorKafkaMessageIterator.java
HenryCaiHaiying: Partition rebalance also happen in the old days for the old consumers, in that case, we just let it continue.  In the end, only one worker is able to upload/persist the datafile for a given partition because each upload has to make sure the pre-condition satisfied (starting offset on S3 for the given partition matches with its own data structure)

In the new consumer case, do you store the offsets in dual storage (both ZK and kafka topic)?  What's the value of offsets.storage and dual.commit.enabled parameters in your case?  I would think kafka would still maintain the offsets in __consumer_topics during rebalance (eliminate the need to manually call seek) if either offsets.storage was set to kafka or dual committing is turned on.

src/main/java/com/pinterest/secor/tools/TestLogMessageProducer.java
HenryCaiHaiying: sounds good

src/main/java/com/pinterest/secor/reader/SecorKafkaMessageIterator.java
igorcalabria: Currently, I'm ignoring both `offsets.storage` and `dual.commit.enabled` but I'll fix it. If `offset.storage` is set to kafka I won't manually seek the offsets. I just need to add a config check so that the user does not set `offset.storage` to kafka and `dual.commit.enabled` to false.

I thought that I wasn't going to be able to commit offsets to kafka, but I was wrong. Secor always commits to zookeeper and this prevents the clients from uploading duplicate offsets. Even if zookeeper commit fails, the second upload should overwrite the first because the filename will be same. Is this correct or have I missed something?

HenryCaiHaiying:
On the uploading part, secor was using the S3 filename (which contains the
previous committed offset) to achieve exact-once atomicity, so it doesn't
matter which worker gets to upload first (or upload the same thing twice).

For offset.storage part, if we use ZK storage mode, the current code should
work (the Uploader.java was the one reading and writing committed
offsets).  If we use kafka storage mode, we probably also need to make a
manual offset commit through kafka consumer in Uploader.java.  By default
we set 'auto.commit.enable' to false for kafka consumer, Uploader.java
probably is the best place to do a manual commit.

igorcalabria:
@HenryCaiHaiying I've made the requested changes and the CI build is passing for all profiles. Here's a list of notable changes since the last review: 

- New consumer now supports storing offsets in kafka. If offset storage is set to kafka and dual commit is enabled, secor skips reading offsets from zookeeper when a re-balance happens https://github.com/pinterest/secor/pull/450/commits/bfa6b55e79ef24625664fafc3765605856134584
- Restore original behavior when a `ConsumerTimeout` happens on the legacy consumer. I was wrongly returning `false` in `hasNext` when this happened https://github.com/pinterest/secor/pull/450/commits/7f49e94470411e4f4524122013904f9a5819cdd5

src/main/java/com/pinterest/secor/reader/SecorKafkaMessageIterator.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    private Map<TopicPartition, Long> getCommittedOffsets(Collection<TopicPartition> assignment) {
+        Map<TopicPartition, Long> committedOffsets = new HashMap<>();
+
+        for (TopicPartition topicPartition : assignment) {
+            com.pinterest.secor.common.TopicPartition secorTopicPartition =
+                    new com.pinterest.secor.common.TopicPartition(topicPartition.topic(), topicPartition.partition());
+            try {
+                long committedOffset = mZookeeperConnector.getCommittedOffsetCount(secorTopicPartition);
```
HenryCaiHaiying: Does this code need to change if the user is not using ZK to store kafka offsets?

igorcalabria: No, because secor skips calling this code if `dual.commit.enabled=true` and `offset.storage=kafka` https://github.com/pinterest/secor/pull/450/files#diff-c2752b6d82af124c79e96142788c189dR104 . `dual.commit.enabled=false` and `offset.storage=kafka` combination was ignored prior to this PR and I've kept the same behavior

igorcalabria:
Do we have the end-to-end test coverage on:

- dual.commit.enabled = true
 Yes, this is covered here https://github.com/pinterest/secor/pull/450/files#diff-bc7e5afea6fc9d87cd71b828456bd6c1R323
- dual.commit.enabled = false, offsets.storage = ZK
Kinda, this is the default configuration set by the properties file, but there's no test that convers a secor commit, a restart and validating that it started from that point.
- dual.commit.enabled = false, offsets.storage = kafka?
No, there's no integration test for this configuration. I'll add one.

For the old consumer, we probably need to have an E2E test coverage on:
- dual.commit.enabled = false, offsets.storage = ZK
This is the default configuration. But it falls in the same situation as the new consumer, there wasn't a test that includes a secor commit and verifying that it starts from that point after a restart. 

I could replicate this test https://github.com/pinterest/secor/pull/450/files#diff-bc7e5afea6fc9d87cd71b828456bd6c1R323 for the configurations you have mentioned since we really should be testing a commit -> restart cycle for all of them.

ambhojgoyal:
@igorcalabria @HenryCaiHaiying When we can have this PR in master ?
Waiting for this integration for long time.

ambhojgoyal:
@igorcalabria @HenryCaiHaiying When we can have this PR in master ?
Waiting for this integration for long time.

Eugst:
+1 
looking forward for this PR merged. 
any ETA?
TIA!

Eugst:
+1 
looking forward for this PR merged. 
any ETA?
TIA!

Eugst:
+1 
looking forward for this PR merged. 
any ETA?
TIA!

Eugst:
+1 
looking forward for this PR merged. 
any ETA?
TIA!

Eugst:
+1 
looking forward for this PR merged. 
any ETA?
TIA!

igorcalabria:
Sorry I took so long. I have updated the PR with the missing testing cases, but one of them is failing for 0.8 profile. I should have time to look into it tomorrow

HenryCaiHaiying:
@igorcalabria thanks for putting in the extra tests.  Looks like there is one test failure in kafka 0.8.2:

2018-10-23 21:45:36,004 [Thread-3] (com.pinterest.secor.main.ConsumerMain$1:65) ERROR Thread Thread[Thread-3,5,main] failed
java.lang.RuntimeException: Failed to initialize the consumer
	at com.pinterest.secor.consumer.Consumer.run(Consumer.java:90)
Caused by: kafka.common.ConsumerRebalanceFailedException: secor_partition_travis-job-9809d975-8fd2-4181-9647-acc606f2ec93_10159_12 can't rebalance after 4 retries
	at kafka.consumer.ZookeeperConsumerConnector$ZKRebalancerListener.syncedRebalance(ZookeeperConsumerConnector.scala:633)
	at kafka.consumer.ZookeeperConsumerConnector.kafka$consumer$ZookeeperConsumerConnector$$reinitializeConsumer(ZookeeperConsumerConnector.scala:905)

Eugst:
I'm sorry to bother you... :)
any ETA?

igorcalabria:
Sorry about the delay, I'll have time to fix the tests this week

HenryCaiHaiying:
Thanks Igor.

src/main/java/com/pinterest/secor/reader/SecorKafkaMessageIterator.java
```diff
@@ -0,0 +1,166 @@
...
+                    } else {
+                        LOG.debug("Seeking {} to offset {}", topicPartition, Math.max(0, offset));
+                        mKafkaConsumer.seek(topicPartition, Math.max(0, offset));
+                    }
+                }));
+            }
+        };
+
+        String[] subscribeList = config.getKafkaTopicList();
+        if (Strings.isNullOrEmpty(subscribeList[0])) {
```
Faqa: Sorry to weigh in while not being a member of the secor project, but I AM running this branch in order to work with a Kafka cluster that includes authentication. This will fail if the user has not specified a topic list.

HenryCaiHaiying: Why the user does not have a topic list?

Faqa:
Because they're subscribing based only on topic patterns.

src/main/java/com/pinterest/secor/reader/SecorKafkaMessageIterator.java
Faqa: Because they're subscribing based only on topic patterns.

juhoautio:
@igorcalabria, does this PR support seamlessly migrating from zk to kafka as the offset storage without losing any data? Sorry, I read the comments in the PR but based on them I wouldn't be sure. Does it make secor initially read the offsets from zk but then switch to kafka? How does secor know to do that? What is the migration process like – do I need to set some specific conf, should I first make sure that all instances are stopped, etc.? I'd be willing to test this PR as long as I know how :)

(I posted essentially the same question in this issue: https://github.com/pinterest/secor/issues/447#issuecomment-434801906)

igorcalabria:
@juhoautio Yeah, you should be able to migrate offsets from zookeeper to kafka. I just read your issue and there may be a bug when the old consumer calls `commit()`. It calls commit without arguments and I'm pretty sure that it commits all consumed messages [[1]](https://github.com/apache/kafka/blob/0.10.2/core/src/main/scala/kafka/consumer/ZookeeperConsumerConnector.scala#L321) instead of just the ones that were uploaded. 

It would be great if you tested offset migration with this version. There are two steps: 
- You have to restart your consumers setting `offset.storage=kafka` but leaving `dual.commit.enabled=false`. This means that secor will also commit uploaded offsets to kafka but it'll keep using zookeeper to seek those offsets. This sounds really confusing(it is!) but I've tried to keep the old code intact. 

- When you confirm that all desired offsets were also commited to kafka, you can change `dual.commit.enabled` to `true` and restart your consumers. When `dual.commit.enabled=true` and `offset.storage=kafka` this new version seeks offsets using kafka storage but keeps using zookeeper for the upload checks

src/main/java/com/pinterest/secor/reader/SecorKafkaMessageIterator.java
igorcalabria: This shouldn't break since `kafka.new.consumer.topic.list` is set on `secor.common.properties`. Are you using kubernetes? I forgot to add a bunch configs to the other properties files since they aren't tested.

Faqa: Yeah, this turned out to be an error on my end. Apologies, ignore.

jutley:
Does this PR allow for Secor to be configured to not use Zookeeper at all? I've tried running this PR with many different configurations, but every configuration seems to write to Zookeeper. Perhaps I'm missing something?

Kafka clients writing directly to Zookeeper is an established anti-pattern. If this can be eliminated from Secor, that would be a big win, imo.

Eugst:
:crying_cat_face: no news so far?

Eugst:
Thanks a lot @igorcalabria !

@HenryCaiHaiying looking forward for your review

norrs:
Nice contribution @igorcalabria  :+1:

juhoautio:
@igorcalabria thanks a lot.

Based on an initial test everything seems fine. For us it may be enough to just have the `offset.storage=kafka` enabled so that we can track offsets via kafka offset storage. Is there any problem with that?

Are there any other reasons to migrate further to `dual.commit.enabled=true`, or is it just that with 2.0.0 that will somehow be the only option? I'm a bit confused because I thought with both flags flipped Secor still commits offsets to both storages, so there's no config combination that would make Secor commit offsets _only_ to kafka?

jutley:
@juhoautio From what I'm reading in this PR, it seems like Secor is still writing offsets to Zookeeper. It needs to keep track of offsets independently from Kafka's internal offsets topic. Tools like Kafka Connect do this by writing to their own special Kafka topics, but Secor seems to still use Zookeeper.

Hopefully this can get addressed in the future, as it's the main reason I'm shying away from using Secor.

juhoautio:
Actually no. I didn't realize that the fixes only target the kafka-2.0.0 profile. Why so?

We use kafka-1.0.0 profile where SecorKafkaMessageIterator is not available. It still has the same bug for kafka commit, it ignores the partition & offset:
https://github.com/pinterest/secor/blob/b97bfded179e0a99b907eb989b9e0e3deaed5b40/src/main/java/com/pinterest/secor/reader/LegacyKafkaMessageIterator.java#L77

I don't think we're upgrading to 2.0.0 very soon. But why not have at least the offset commit fix also in LegacyKafkaMessageIterator? Were there any other crucial fixes for the dual commit setup?

src/main/java/com/pinterest/secor/reader/SecorKafkaMessageIterator.java
```diff
@@ -0,0 +1,166 @@
...
+    @Override
+    public void commit(com.pinterest.secor.common.TopicPartition topicPartition, long offset) {
+        TopicPartition kafkaTopicPartition = new TopicPartition(topicPartition.getTopic(), topicPartition.getPartition());
+        OffsetAndMetadata offsetAndMetadata = new OffsetAndMetadata(offset);
+
+        try {
+            LOG.info("committing {} offset {} to kafka", topicPartition, offset);
+            mKafkaConsumer.commitSync(ImmutableMap.of(kafkaTopicPartition, offsetAndMetadata));
+        } catch (CommitFailedException e) {
+            LOG.trace("kafka commit failed due to group re-balance", e);
```
ujjwal: Should this exception be rethrown? What if it can't commit the offset?

ujjwal: https://github.com/pinterest/secor/blob/master/src/main/java/com/pinterest/secor/reader/LegacyKafkaMessageIterator.java#L77

`LegacyKafkaMessageIterator` - doesn't swallow this exception.

igorcalabria: I'm pretty sure that we don't need to re-throw it. At this point, the offset was committed to zookeeper and the partition is assigned to another consumer instance. If the consumer is configured to read offsets from kafka, it'll read already uploaded offsets but they'll be skipped since zookeeper commit was successful. If the consumer was configured to read offsets from zookeeper, it will simply continue from there.  

`LegacyKafkaMessageIterator` does not swallow it because it's another lib. I'm not that familiar with the old API, but I believe that it does not have an equivalent exception to `CommitFailedException`. Unfortunately, the exception name can be very misleading since this is not a generic error. It only happens when the commit fails and cannot be retried (group rebalance).

src/main/java/com/pinterest/secor/reader/SecorKafkaMessageIterator.java
glasser: Or maybe I'm just confused about the overall model here.  If I'm currently using the legacy iterator (or well, an older version Secor) and zookeeper offset storage, what's the right way to upgrade to the using the newer iterator? Should I do a two-step upgrade where I first upgrade to "legacy iterator, kafka storage, dual commit", then to "secor iterator, kafka storage, dual commit"?  Or a three-step upgrade going to "legacy, kafka, dual" -> "legacy, kafka, no dual" -> "secor, kafka, no dual"?  Or just a one-step upgrade?

glasser:
I'm a little surprised this PR doesn't let you adjust `session.timeout.ms` from Kafka's default of 10s.  If I understand the semantics of the new consumer correctly, this means that if you don't call poll at least every 10 seconds, the group coordinator will give up on the consumer — ie, if uploading a partition ever takes more than 10 seconds, rebalancing will happen. Has nobody else run into this?

(Though if we do increase that amount we should probably also make sure to call `close()` when the consumer thread ends, especially if you merge #481.)

glasser:
Is anybody successfully using this Kafka 2.0.0 client in production? Curious about my questions at https://github.com/pinterest/secor/pull/450#issuecomment-450680633 and https://github.com/pinterest/secor/pull/450#discussion_r244368394

juhoautio:
@glasser yes, using in production but sorry I can’t really comment on the settings that you asked about. We use the settings where secor commits to both offset storages.

karishmakhubchandani:
Hi @glasser  which version of secor you are using ? and which version of kafka client it uses ? 
is secor by default adding kafka-client 2.0.0 or do we have to manually add kafka 2.0.0 client lib in secor ? 
currently I am using secor 0.21 and it uses kafka-client 0.8.2.1 lib.

glasser:
I've been building from master, which is a fair amount newer than 0.21.

PR #451, opened 2018-09-01
fix: Set default locale to US

yujhe:
The date time format is for Locale.US.


HenryCaiHaiying:
Thanks for the contribution.

PR #454, opened 2018-09-11
Avoid resolving ZK hosts too early

spektom:
What happens is during the construction of ZK client provided addresses are resolved due to default behaviour of InetSocketAddress. Now, if a ZK server changes its IP address Secor stops working until restarted. Here are the error that we get:

```
[Thread-15-SendThread(ip-10-118-20-99.eu-west-1.compute.internal:2181)] WARN org.apache.zookeeper.ClientCnxn - Session 0x100a9194f500006 for server null, unexpected error, closing socket connection and attempting reconnect
java.net.NoRouteToHostException: Host is unreachable
at sun.nio.ch.SocketChannelImpl.checkConnect(Native Method)
at sun.nio.ch.SocketChannelImpl.finishConnect(SocketChannelImpl.java:717)
at org.apache.zookeeper.ClientCnxnSocketNIO.doTransport(ClientCnxnSocketNIO.java:361)
at org.apache.zookeeper.ClientCnxn$SendThread.run(ClientCnxn.java:1081)
```

### DIFF
--- a/src/main/java/com/pinterest/secor/common/ZookeeperConnector.java
+++ b/src/main/java/com/pinterest/secor/common/ZookeeperConnector.java
@@ -67,7 +67,7 @@ public class ZookeeperConnector {
             assert elements.length == 2: Integer.toString(elements.length) + " == 2";
             String host = elements[0];
             int port = Integer.parseInt(elements[1]);
-            result.add(new InetSocketAddress(host, port));
+            result.add(InetSocketAddress.createUnresolved(host, port));
         }
         return result;
     }


src/main/java/com/pinterest/secor/common/ZookeeperConnector.java
```diff
@@ -67,7 +67,7 @@ public ZookeeperConnector(SecorConfig config) {
             assert elements.length == 2: Integer.toString(elements.length) + " == 2";
             String host = elements[0];
             int port = Integer.parseInt(elements[1]);
-            result.add(new InetSocketAddress(host, port));
+            result.add(InetSocketAddress.createUnresolved(host, port));
```
HenryCaiHaiying: I am not quite familiar with this domain, are you using a proxy?  If the address is not resolved, when will it resolve?  And if the ip address of the ZK changed, will this unresolved address re-resolve?

spektom:
Let's say I'm using a ZK hostnames (not IP addresses) when starting Secor.
After some time, one of died ZK nodes is replaced, and the hostname now
points to a new node. If using the original constructor of
InetSocketAddress the client will never know about the change because the
hostname was resolved to IP address early (in constructor). Unresolved
version of this class delays resolution of a hostname until it's used.

HenryCaiHaiying:
If the hostname -> ipaddress was resolved early when its first used.  And then the ipaddress changed for that host, how does the InetSocketAddress class knows it needs to be re-resolve?

spektom:
If using createUnresolved() instead of regular constructor, hostname is
kept unresolved. It's only resolved once it's used (for example when
connecting to ZK).

If the IP changes, the ZK client will get a timeout, and the hostname will
be resolved to the new IP during reconnect.

HenryCaiHaiying:
Where (which class/method) did the following?

If the IP changes, the ZK client will get a timeout, and the hostname will
be resolved to the new IP during reconnect.

spektom:
Look at this class:
https://github.com/twitter/commons/blob/master/src/java/com/twitter/common/zookeeper/ZooKeeperClient.java

In method get(...):

"Returns the current active ZK connection or establishes a new one if none
has yet been
established or a previous connection was disconnected or had its session
time out...."

As long as this class is initialized with hostname and not with IP address
it will reconnect using up to date ZK servers.

HenryCaiHaiying:
Looks good, I will merge your PR

PR #459, opened 2018-11-08
POC: Commit offsets also to kafka while keeping offsets.storage=zookeeper

juhoautio:
Secor is still heavily dependent on using Zookeeper as an offset storage:
- There doesn't seem to be a configuration combination for kafka version profiles < 2.0.0 that would allow committing offsets to kafka without loss of data.
   - If `kafka.offsets.storage=kafka`, offsets of all partitions are committed at once (see `LegacyKafkaMessageIterator#commit`), which can cause data loss
- Even with the 2.0.0 profile offsets are committed to kafka only if `kafka.offsets.storage=kafka`

----

My use case is that there are many kafka offset monitoring tools that only support the kafka offset storage, not zookeeper. And we're not likely to upgrade to kafka 2.0.0 very soon.

To support that secor would have to either:
**A**) support committing offsets additionally to kafka offset storage
**B**) support committing offsets correctly in `LegacyKafkaMessageIterator#commit`

This PR is an _incremental_ POC of **A** – adding this doesn't change any of the existing behavior, so at least this is totally safe.

However, it would be better if **B** can be done. Is there any reason why `LegacyKafkaMessageIterator#commit` couldn't be changed to do what `#commitToKafka` in this PR does? If not, then it would be enough to implement **B**.

**TODO:** (for **B**) I'm not sure if the `ConsumerConnector#commitOffsets` (that `LegacyKafkaMessageIterator` uses) is synchronous or not – but would it even have to be? I'm wondering because `SecorKafkaMessageIterator` calls `KafkaConsumer#commitSync`.

----

For background: https://github.com/pinterest/secor/pull/450#issuecomment-436914329 and also the PR https://github.com/pinterest/secor/pull/450 as a whole.

----

Travis build currently fails for `MVN_PROFILE=kafka-0.8.2.1`. This will be fixed if this PR gets some yellow light.

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:2.0.2:compile (default-compile) on project secor: Compilation failure: Compilation failure: 
[ERROR] /home/travis/build/pinterest/secor/src/main/java/com/pinterest/secor/reader/LegacyKafkaMessageIterator.java:[10,19] error: cannot find symbol
[ERROR] 
[ERROR] could not parse error message:   symbol:   class OffsetMetadata
[ERROR]   location: package kafka.common
[ERROR] /home/travis/build/pinterest/secor/src/main/java/com/pinterest/secor/reader/LegacyKafkaMessageIterator.java:15: error: cannot find symbol
[ERROR] import org.apache.kafka.clients.consumer.CommitFailedException;
[ERROR]                                         ^
[ERROR] 
[ERROR] could not parse error message:   symbol:   class CommitFailedException
[ERROR]   location: package org.apache.kafka.clients.consumer
[ERROR] /home/travis/build/pinterest/secor/src/main/java/com/pinterest/secor/reader/LegacyKafkaMessageIterator.java:94: error: cannot find symbol
[ERROR]         OffsetAndMetadata offsetAndMetadata = new OffsetAndMetadata(new OffsetMetadata(offset, null),
[ERROR]                                                                         ^
[ERROR] 
[ERROR] could not parse error message:   symbol:   class OffsetMetadata
[ERROR]   location: class LegacyKafkaMessageIterator
[ERROR] /home/travis/build/pinterest/secor/src/main/java/com/pinterest/secor/reader/LegacyKafkaMessageIterator.java:100: error: no suitable method found for commitOffsets(Map<TopicAndPartition,OffsetAndMetadata>,boolean)
[ERROR]             mKafkaCommitter.commitOffsets(offsets, true);
[ERROR]                            ^
[ERROR] 
[ERROR]     method ConsumerConnector.commitOffsets() is not applicable
[ERROR]       (actual and formal argument lists differ in length)
[ERROR]     method ConsumerConnector.commitOffsets(boolean) is not applicable
[ERROR]       (actual and formal argument lists differ in length)
[ERROR] /home/travis/build/pinterest/secor/src/main/java/com/pinterest/secor/reader/LegacyKafkaMessageIterator.java:[101,17] error: cannot find symbol
```

### DIFF
--- a/src/main/config/secor.common.properties
+++ b/src/main/config/secor.common.properties
@@ -231,6 +231,12 @@ kafka.zookeeper.path=/
 #Only acquired used for decoding Avro messages
 schema.registry.url=
 
+# Commit offsets additionally to kafka offsets storage (for reference).
+# This is just to allow tracking offsets also via Kafka offset storage. Secor still uses Zookeeper as the offset storage
+# internally.
+# Possible values: true or false - true is not allowed if kafka.offsets.storage=kafka
+secor.offsets.commitToKafkaOffsetStorageForReference=false
+
 # Store offset in zookeeper and kafka consumer topic.
 # Only used if kafka.offsets.storage is set to "kafka"
 # http://kafka.apache.org/documentation.html#oldconsumerconfigs
--- a/src/main/java/com/pinterest/secor/common/SecorConstants.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConstants.java
@@ -3,4 +3,6 @@ package com.pinterest.secor.common;
 public class SecorConstants {
     public static final String KAFKA_OFFSETS_STORAGE_ZK = "zookeeper";
     public static final String KAFKA_OFFSETS_STORAGE_KAFKA = "kafka";
+    public static final String COMMIT_TO_KAFKA_OFFSET_STORAGE_FOR_REFERENCE =
+            "secor.offsets.commitToKafkaOffsetStorageForReference";
 }
--- a/src/main/java/com/pinterest/secor/reader/KafkaMessageIterator.java
+++ b/src/main/java/com/pinterest/secor/reader/KafkaMessageIterator.java
@@ -11,4 +11,5 @@ public interface KafkaMessageIterator {
     Message next();
     void init(SecorConfig config) throws UnknownHostException;
     void commit(TopicPartition topicPartition, long offset);
+    void commitToKafka(TopicPartition topicPartition, long offset);
 }
--- a/src/main/java/com/pinterest/secor/reader/LegacyKafkaMessageIterator.java
+++ b/src/main/java/com/pinterest/secor/reader/LegacyKafkaMessageIterator.java
@@ -1,18 +1,25 @@
 package com.pinterest.secor.reader;
 
+import com.google.common.collect.ImmutableMap;
 import com.pinterest.secor.common.SecorConfig;
 import com.pinterest.secor.common.TopicPartition;
 import com.pinterest.secor.message.Message;
 import com.pinterest.secor.timestamp.KafkaMessageTimestampFactory;
 import com.pinterest.secor.util.IdUtil;
+import kafka.common.OffsetAndMetadata;
+import kafka.common.OffsetMetadata;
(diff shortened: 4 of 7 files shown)


juhoautio:
@igorcalabria could you comment on this please:

> Is there any reason why LegacyKafkaMessageIterator#commit couldn't be changed to do what #commitToKafka in this PR does?

igorcalabria:
> @igorcalabria could you comment on this please:
> 
> > Is there any reason why LegacyKafkaMessageIterator#commit couldn't be changed to do what #commitToKafka in this PR does?

No reason I can think off. I didn't add the fix to the 2.0 PR because it was big enough and we didn't have any test cases that cover the failure state caused by the kafka commit on the legacy consumer.

Oh, and I noticed that you said that you aren't upgrading to kafka-2.0.0 anytime soon. From broker version 0.10 (I'm not sure on the exact version) onwards you can use a new consumer with an older broker without problems. We are using 2.0 consumers with 1.0 brokers. The performance penalty is only the other way around, older consumers with new brokers(when there's protocol change)

juhoautio:
Thanks @igorcalabria!

> I also saw your comments about kafka only commit. Maybe you could open an issue and start the discussion there, it's not the first I time I've seen this feature requested.

I'm not sure what you're referring to as "kafka only commit" in what I have written?

igorcalabria:
> Thanks @igorcalabria!
> 
> > I also saw your comments about kafka only commit. Maybe you could open an issue and start the discussion there, it's not the first I time I've seen this feature requested.
> 
> I'm not sure what you're referring to as "kafka only commit" in what I have written?

I must have mixed something up. It was another person talking about migrating the commit logic to only use kafka. Sorry.

juhoautio:
Closing this because we were able to instead upgrade to using the kafka-2.0.0 build profile of secor. As @igorcalabria wrote above, that consumer library 2.0.0 is compatible with our kafka server version even though it's a bit older.

So, nothing wrong with this PR per se, but maybe it's simpler to keep secor as it is. If anyone needs those @igorcalabria's fixes, they should be able to upgrade their kafka server, if still running a version that's so old that it doesn't work with 2.0.0 consumer.

PR #462, opened 2018-11-09
Disable the use of the system classloader for surefire

ankon:
This works around https://issues.apache.org/jira/browse/SUREFIRE-1588.

---
This fixes building on recent Debians (particularly inside the docker `openjdk:8-jdk` image) for me.


HenryCaiHaiying:
Looks like it has conflict on the latest pom.xml, can you take a look?

PR #466, opened 2018-11-12
Fix typo: ORCScehmaProvider -> ORCSchemaProvider

ankon:


ankon:
Hmm. This seems to be a typo existing in the source as well? https://github.com/pinterest/secor/blob/master/src/main/java/com/pinterest/secor/util/orc/schema/ORCScehmaProvider.java

PR #468, opened 2018-11-15
Dump the final secor configuration on the 'debug' log level

### LINKED ISSUE #465: Dumping secor configuration at startup
I'm still trying to configure secor, and what helped me a bit was to add this change to SecorConfig so it would dump the full configuration at startup, after all `include` and system properties have been applied:

```diff
diff --git a/src/main/java/com/pinterest/secor/common/SecorConfig.java b/src/main/java/com/pinterest/secor/common/SecorConfig.java
index 3c55ea6..4839ac7 100644
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -18,8 +18,11 @@ package com.pinterest.secor.common;
 
 import com.google.api.client.repackaged.com.google.common.base.Strings;
 import org.apache.commons.configuration.ConfigurationException;
+import org.apache.commons.configuration.ConfigurationUtils;
 import org.apache.commons.configuration.PropertiesConfiguration;
 import org.apache.commons.lang.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.HashMap;
 import java.util.Iterator;
@@ -33,6 +36,8 @@ import java.util.TimeZone;
  * @author Pawel Garbacki (pawel@pinterest.com)
  */
 public class SecorConfig {
+    private static final Logger LOG = LoggerFactory.getLogger(SecorConfig.class);
+
     private final PropertiesConfiguration mProperties;
 
     private static final ThreadLocal<SecorConfig> mSecorConfig = new ThreadLocal<SecorConfig>() {
@@ -54,6 +59,7 @@ public class SecorConfig {
                 properties.setProperty(entry.getKey().toString(), entry.getValue());
             }
 
+            LOG.trace("Configuration: " + ConfigurationUtils.toString(properties));
             return new SecorConfig(properties);
         }
     };
```

Would this be something acceptable, maybe in a cleaned up/nicer formatted way?

ankon:
Fixes #465

### DIFF
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -18,8 +18,11 @@ package com.pinterest.secor.common;
 
 import com.google.api.client.repackaged.com.google.common.base.Strings;
 import org.apache.commons.configuration.ConfigurationException;
+import org.apache.commons.configuration.ConfigurationUtils;
 import org.apache.commons.configuration.PropertiesConfiguration;
 import org.apache.commons.lang.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.HashMap;
 import java.util.Iterator;
@@ -33,6 +36,8 @@ import java.util.TimeZone;
  * @author Pawel Garbacki (pawel@pinterest.com)
  */
 public class SecorConfig {
+    private static final Logger LOG = LoggerFactory.getLogger(SecorConfig.class);
+
     private final PropertiesConfiguration mProperties;
 
     private static final ThreadLocal<SecorConfig> mSecorConfig = new ThreadLocal<SecorConfig>() {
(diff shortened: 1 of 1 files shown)


src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -54,6 +59,7 @@ protected SecorConfig initialValue() {
                 properties.setProperty(entry.getKey().toString(), entry.getValue());
             }
 
+            LOG.debug("Configuration: " + ConfigurationUtils.toString(properties));
```
HenryCaiHaiying: This '+' operation might be expensive, does this scheme work?

LOG.debug("Configuration: {}", properties)

If not, you can do:
    if (LOG.isDebugEnabled()) {
        LOG.debug ...
    }

ankon: Good point. I'll try, and update the branch then.

ankon:
Actually I think it made sense to apply _both_ suggestions:
- Using the template string looks cleaner than the string concatenation, and seems to be a "best practice" for logging
- The most important cost of dumping the configuration isn't in that string concatenation, but rather in building the string in the first place. Now this code is only run once, but it still seems prudent to avoid wasting cycles here.

PR #475, opened 2018-12-04
Don't depend on nested repackaged Guava

glasser:
We have a perfectly fine copy of com.google.common.base.Strings that doesn't
depend on implementation details of the Google API client.


glasser:
This caused a problem for me in production when trying to use an upgraded version of the Google API client locally.

HenryCaiHaiying:
Looks good.

PR #479, opened 2018-12-29
ProgressMonitor: support Datadog-style tags

glasser:
Datadog's java-dogstatsd-client is a backwards-compatible extension to
java-statsd-client which works with normal statsd servers as well as Datadog's
dogstatsd server. This change is backwards compatible.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -334,9 +334,9 @@
             </exclusions>
         </dependency>
         <dependency>
-            <groupId>com.timgroup</groupId>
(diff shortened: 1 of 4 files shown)


glasser:
Here's an example of another project we use making a similar change (just to demonstrate that this library swap isn't completely unreasonable): https://github.com/apache/incubator-druid/pull/6605

HenryCaiHaiying:
Looks good.

PR #480, opened 2018-12-29
Avoid putting Message key/payload in an exception

### LINKED ISSUE #474: Remove or truncate payload and kafkaKey in Message.fieldsToString()
Sometimes, a stringified Message or ParsedMessage ends up in, say, a RuntimeException that is thrown and printed at crash time.  These classes contain kafkaKey and (especially) payload which can be very long (kilobytes or megabytes in our use case), multiple lines, contain PII, etc. I'd like to either truncate these fields to a much smaller length (say 1K) or even leave the fields out entirely. Does anything depend on these fields being fully included in toString?

glasser:
This exception happens frequently due to IO errors or due to FileRegistry
throwing during shutdown, so including what might be a ginormous binary blob
isn't great.

While we're at it, ensure that Message.toString() doesn't throw on non-UTF-8
data by explicitly using CharsetDecoder rather than relying on explicitly
unspecified behavior of `new String(byte[])`.

Fixes #474.


HenryCaiHaiying:
Looks OK.  For the code commenting style, once you got multiple lines in comments, it's better use the /* ... */ block style comment rather than one liner comment //

glasser:
Adjusted comments, thanks.

PR #481, opened 2018-12-29
Upload during graceful shutdown

glasser:
This new feature is disabled by default. However, there are two
other small changes introduced even with secor.upload.on.shutdown=false:

- If KafkaMessageIterator returns false from hasNext (only possible with the
  legacy iterator), perform a final upload at the end in this case too (whether
  or not the graceful shutdown feature is enabled).

- After trimming files, we immediate check again to see if we're supposed to
  upload the TopicPartition. This is needed for graceful shutdown (since there's
  not going to be another call to checkTopicPartition) and may also be helpful
  in other configurations.

### DIFF
--- a/src/main/java/com/pinterest/secor/consumer/Consumer.java
+++ b/src/main/java/com/pinterest/secor/consumer/Consumer.java
@@ -59,6 +59,10 @@ public class Consumer extends Thread {
     protected Uploader mUploader;
     // TODO(pawel): we should keep a count per topic partition.
     protected double mUnparsableMessages;
+    // If we aren't configured to upload on shutdown, then don't bother to check
+    // the volatile variable.
+    private boolean mUploadOnShutdown;
+    private volatile boolean mShuttingDown = false;
 
     public Consumer(SecorConfig config) {
         mConfig = config;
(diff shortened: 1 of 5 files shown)


src/main/java/com/pinterest/secor/consumer/Consumer.java
```diff
@@ -78,6 +82,24 @@ private void init() throws Exception {
...
         mUnparsableMessages = 0.;
+
+        mUploadOnShutdown = mConfig.getUploadOnShutdown();
+        if (mUploadOnShutdown) {
+            Runtime.getRuntime().addShutdownHook(this.new ShutdownHook());
+        }
+    }
+
+    // When the JVM starts to shut down, tell the Consumer thread to upload once and wait for it to finish.
+    private class ShutdownHook extends Thread {
```
HenryCaiHaiying: Maybe the class name should be called FinalUploadShutdownHook to make the purpose more explicit

glasser: Done

glasser:
Oops! This feature creates a deadlock.

If one of the consumer threads throws an unhandled exception, the UncaughtExceptionHandler in ConsumerMain will call System.exit(1).

**It calls this synchronously from within the thread itself.** That is, the consumer thread is now blocked on waiting for shutdown hooks to finish.

But the registered shutdown hook blocks on waiting for the consumer thread to finish. Oops! Deadlock — the rest of the program keeps running, though probably the other consumer threads will fall into the same deadlock sooner or later, and then you just have a zombie program.

I'll submit a PR to fix it.

glasser:
See #593.

PR #567, opened 2019-02-15
Improve license on files header

drrzmr:
- Use license text from: https://www.apache.org/legal/src-headers.html#headers
- Remove javadoc dangling comments
- Add license to all java files


HenryCaiHaiying:
Looks good, thanks.

PR #568, opened 2019-02-16
added option to change S3Uploader for FakeS3 when running with SECOR_…

nikshepsvn:
…DEV='true'

### DIFF
--- a/src/main/java/com/pinterest/secor/uploader/S3UploadManager.java
+++ b/src/main/java/com/pinterest/secor/uploader/S3UploadManager.java
@@ -128,6 +128,9 @@ public class S3UploadManager extends UploadManager {
         if (mConfig.getAwsClientPathStyleAccess()) {
             S3ClientOptions clientOptions = new S3ClientOptions();
             clientOptions.setPathStyleAccess(true);
+            if (System.getenv("SECOR_DEV") == "true"){
+                clientOptions.setChunkedEncodingDisabled();
+            }
             client.setS3ClientOptions(clientOptions);
         }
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/pinterest/secor/uploader/S3UploadManager.java
```diff
@@ -128,6 +128,9 @@ public void refresh() {}
         if (mConfig.getAwsClientPathStyleAccess()) {
             S3ClientOptions clientOptions = new S3ClientOptions();
             clientOptions.setPathStyleAccess(true);
+            if (System.getenv("SECOR_DEV") == "true"){
+                clientOptions.disableChunkedEncoding();
```
HenryCaiHaiying: What was the problem with checked encoding?

HenryCaiHaiying: And when is SECOR_DEV being set?

drrzmr: I guess that @nikshepsvn are setting `SECOR_DEV` only in his development environment. IMHO this approach are like "easter eggs", better approach are use properties to change this, where it must be well documented what is the propose of this setting.

nikshepsvn:
Oops everyone, was supposed to make this PR on a fork of this repo! What I've heard is that fakeS3 bugs out when using chunkedEncoding (see: https://github.com/jubos/fake-s3/issues/164) but even after I figured that out and got it working there was some SSL issues connecting to the bucket which I have created issues for.

PR #593, opened 2019-02-27
Fix deadlock when secor.upload.on.shutdown=true

glasser:
The secor.upload.on.shutdown feature worked fine if we were shutting down
because we received a signal, but not if we were shutting down because a
Consumer thread had crashed and the UncaughtExceptionHandler had called
System.exit(1). In that case we'd get a deadlock and the program would never
exit.

Change the shutdown hook to do nothing if any Consumer thread called
System.exit.

The simplest way to implement this is to move the UncaughtExceptionHandler into
the Consumer thread itself as a big try block, so I did that.


glasser:
CI failure is maven.twttr.com being down: https://github.com/twitter/commons/issues/418

HenryCaiHaiying:
Thanks for fixing the twitter maven issue.

Issue #600, opened 2019-02-28
Messages can be uploaded multiple times to different files


glasser:
I think there's a fundamental problem with the Secor upload algorithm that can lead to messages appearing multiple times in backups.

Let's say you have a consumer running on a single topic/partition, dividing up messages into hourly files.  The current committed offset is 11111111111111111111. The messages it finds span an hour break. It has created these two files:

```
y=2019/m=02/d=28/H=10/1_1_11111111111111111111.gz
y=2019/m=02/d=28/H=11/1_1_11111111111111111111.gz
```

It decides to upload, takes out the Zookeeper lock, uploads both files, but crashes before updating the Zookeeper committed offset. (In practice this could be "uploads the first few of a large number of files", especially if you're using something like SplitByFieldMessageParser that produces many files not just on hour boundaries.)

A new consumer takes over this topic/partition and starts writing messages locally. It will choose the same filenames for its files. But for whatever reason, this consumer decides it's time to upload faster than the previous iteration did, before it even gets to the second hour worth of messages. It uploads only this file:

```
y=2019/m=02/d=28/H=10/1_1_11111111111111111111.gz
```

It successfully updates the ZK offset to 11111111111111199999, and continues reading messages.  Now it creates these two files locally and uploads them:

```
y=2019/m=02/d=28/H=10/1_1_11111111111111199999.gz
y=2019/m=02/d=28/H=11/1_1_11111111111111199999.gz
```

Now these files all exist in storage:

```
y=2019/m=02/d=28/H=10/1_1_11111111111111111111.gz (overwritten by the second attempt)
y=2019/m=02/d=28/H=10/1_1_11111111111111199999.gz
y=2019/m=02/d=28/H=11/1_1_11111111111111111111.gz (uploaded by the first attempt)
y=2019/m=02/d=28/H=11/1_1_11111111111111199999.gz
```

But messages in the third file also appear in the fourth file!

I haven't tried making this happen in practice; this is based on reasoning about the code.

My assumption has been that once Secor has made it well past a timestamp range you can assume that every Kafka message has been uploaded exactly once, but this example seems to break that.

One thought for fixing this: Each topic/partition gets a "current upload" node, which is a list of files in storage. The upload logic (Uploader.uploadFiles) becomes:

- Take out lock
- Read current offset from ZK and stop if it doesn't match what we think it should be (no change from present)
- Read "current upload" file if it exists and delete all files named there from storage.
- Write "current upload" file with the list of files we're going to upload to storage.
[...]

PR #602, opened 2019-03-01
Add deterministic upload mode

glasser:
This is to avoid #600.

In this mode, decisions about whether to upload files are *only* based on
properties of the input messages themselves: timestamps and input message
payload size.  We don't care about real-world time, disk file timestamps, or log
file size; we don't support upload on shutdown; and we check for uploads after
every message.

Configuration:

- set secor.upload.deterministic=true
- Configure at least one of secor.max.file.timestamp.range.millis and
  secor.max.input.payload.size.bytes.
- If you've configured secor.max.file.timestamp.range.millis, you must
  set kafka.useTimestamp=true and ensure that your FileReader/FileWriter
  supports timestamps.

### DIFF
--- a/src/main/java/com/pinterest/secor/common/DeterministicUploadPolicyTracker.java
+++ b/src/main/java/com/pinterest/secor/common/DeterministicUploadPolicyTracker.java
@@ -1,0 +1,108 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.pinterest.secor.common;
+
+import com.pinterest.secor.io.KeyValue;
+import com.pinterest.secor.message.Message;
+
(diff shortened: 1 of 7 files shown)


glasser:
I don't know if this fix is good to be upstreamed or if it's what the project wants, but it's what we're going to try for our installation.

src/main/java/com/pinterest/secor/common/DeterministicUploadPolicyTracker.java
```diff
@@ -0,0 +1,109 @@
...
+    if (maxFileTimestampRangeMillis > 0) {
+      mMaxFileTimestampRangeMillis = maxFileTimestampRangeMillis;
+    } else {
+      mMaxFileTimestampRangeMillis = Long.MAX_VALUE;
+    }
+    if (maxInputPayloadSizeBytes > 0) {
+      mMaxInputPayloadSizeBytes = maxInputPayloadSizeBytes;
+    } else {
+      mMaxInputPayloadSizeBytes = Long.MAX_VALUE;
+    }
```
zionts: Should there be validation that at least one value is non-zero?

src/main/java/com/pinterest/secor/consumer/Consumer.java
```diff
@@ -143,14 +156,17 @@ public void run() {
                 }
 
                 long now = System.currentTimeMillis();
-                if (nMessages++ % checkMessagesPerSecond == 0 ||
+                if (mDeterministicUploadPolicyTracker != null ||
```
zionts: Could this still be non-deterministic dependent on the time taken within this loop? Will we be segmenting on disk before the upload (i.e. only uploading the portion that we would have uploaded had we checked in time?)

HenryCaiHaiying:
Looks good to me.  I will merge in this PR, it's a good fix for secor.

jeremyplichtafc:
@glasser - I was reading your code and trying to understand something. If deterministic mode only looks at the time difference in the message for that partition and the size of input for that partition wont you have the case where a partition only gets a handful of messages (never triggering the time or size criteria) and is never uploaded? Especially if you are splitting them up into smaller partitions using SplitByFieldMessageParter.

glasser:
That's probably correct, and should be mentioned in the docs for the option — open a PR?

PR #672, opened 2019-04-14
Send null messages to kafka avro deserializer

shirbr:
Hi Secor community!
We are using Secor for Avro messages with tombstone messaging - meaning we have null values in kafka (for more information: https://debezium.io/blog/tags/smt/#delete_records).
kafka-avro-serializer handles null as payload, so no problem there. The problem we are facing is that Secor converts null values into an empty byte array. In such cases, kafka-avro-serializer is trying to deserialize the data and we get "Error deserializing Avro message" as it is an empty byte array.

We fixed this issue by passing a null value to the serializer, and we now use the serializer results only if they are not null.

We would sincerely appreciate your assistance with this.


HenryCaiHaiying:
Looks fine.

PR #686, opened 2019-04-24
fix messagePack, try-with-resources

sslavian812:
Hi,
I have this small thing fixed during my experiments with Secor.
If some exception happens, the `MessageUnpacker` object will not be closed, so I used try-with resources instead of just closing it manually before `return`.


HenryCaiHaiying:
Thanks for the contribution.

PR #718, opened 2019-05-10
WIP avro field parser

jeremyplichtafc:
Adds the ability to partition by custom fields out of avro messages


HenryCaiHaiying:
Looks like a good addition to secor, are you planning to merge this in (since WIP is still in the title)?

jeremyplichtafc:
Thank you - yes, there were a few more modifications I wanted to make and also wanted to see if it wouldn't be too complicated to add an Avro test to this. I will take the WIP off and ping you when ready

jeremyplichtafc:
I originally stopped work on this because the value I was partioning by had really high cardinality and was causing memory issues. Now that I have another use case that is low cardinality I will be testing out and using this code more and take the WIP off when I feel its stable

PR #881, opened 2019-08-15
changing shell in docker-entrypoint.sh from bash to sh

The-Alchemist:
The base image used in `/Dockerfile`, `openjdk:8` does not contain `/bin/bash`.


HenryCaiHaiying:
Thanks for the contribution.

PR #884, opened 2019-08-19
adding `exec` to the java command in the docker entrypoint

The-Alchemist:
this makes Java PID 1 which is recommended so that SIGINT works properly


HenryCaiHaiying:
Thanks for the contribution.

PR #921, opened 2019-09-06
Add glue metastore update

Sab8:


HenryCaiHaiying:
Are you planning to merge this PR to upstream?  Looks like it's quite a big change which might need some explanation and discussion.

Sab8:
Hi, yes I would like to contribute it. 
The changes include 2 features : 
- A GluePartitionFinalizer (~managed hive metastore on AWS). To reuse code and keep the same logic of the QuboleFinalizer we abstracted it to  AbstractPartitionFinalizer leaving the addPartition method to be implemented by the impl without breaking the default behaviour. 
-  The Glue client required providing a custom credentialsProvider (AssumeRoleAndDefaultCredentialsProviderChain) to be able to do an assume role without breaking the authentication of the S3Uploader.  So we added the ability to override the Credentials Provider with "fs.s3a.aws.credentials.provider" and as AWS regions and enpoints are interchangeable we added the abylity to provide one or the other.

src/main/java/com/pinterest/secor/common/aws/AssumeRoleAndDefaultCredentialsProviderChain.java
```diff
@@ -0,0 +1,72 @@
...
+      WebIdentityTokenCredentialsProvider.builder().build());
+  }
+
+  public static AssumeRoleAndDefaultCredentialsProviderChain getInstance() {
+    return INSTANCE;
+  }
+
+  private static AWSCredentialsProvider getDefaultCredentials() {
+    log.info("Using AssumeRoleAndDefaultCredentialsProviderChain");
+    final String profileName = "profile " + AwsProfileNameLoader.INSTANCE.loadProfileName();
```
holyshipt: when deployed as service this assumes we have awscli setup, I think we can treat it as one of those secret in config?

src/main/java/com/pinterest/secor/finalizer/metastore/GlueClient.java
```diff
@@ -0,0 +1,77 @@
...
+    databaseName = config.getGlueDatabaseName();
+    client = AWSGlueClientBuilder
+      .standard()
+      .withRegion(Regions.fromName(config.getAwsRegion()))
+      .withCredentials(new AssumeRoleAndDefaultCredentialsProviderChain())
+      .withClientConfiguration(new ClientConfiguration().withRequestTimeout(config.getGlueTimeoutMs()))
+      .build();
+  }
+
+  public void addPartition(String table, String partition) {
```
holyshipt: the partition literal was concat'ed by comma previously, here split by '/' will get nothing.
this also doesn't take care of partition by minute? it would have third partition field.

HenryCaiHaiying:
@Sab8 Are you still working on this PR?

Sab8:
No but I can pick it up if there is renewed interest in this.

Le lun. 25 mai 2020 à 21:54, Henry Haiying Cai <notifications@github.com> a
écrit :

HenryCaiHaiying:
I think it would be nice to have this PR merged if you already used in your production environment.  I think you just need to resolve the conflicts with the current secor code base.

Issue #894, opened 2019-08-22
Avro schema not found for topic X 


irajhedayati:
**Related to https://github.com/pinterest/secor/issues/456 but different scenario**

Version: `0.27`

```
Command to run:
sudo docker run  \
    -e ZOOKEEPER_QUORUM='ZK1:2181,ZK2:2181,ZK3:2181' \
    -e ZOOKEEPER_PATH='/' \
    -e AWS_SECRET_KEY='*****' \
    -e AWS_ACCESS_KEY='*****' \
    -e SECOR_GROUP='secor_avro' \
    -e SECOR_S3_BUCKET='*******' \
    -e SECOR_MESSAGE_TIMESTAMP='recordDateTime' \
    -e SECOR_EXTRA_OPTS='-Dsecor.compression.codec=org.apache.hadoop.io.compress.GzipCodec -Dsecor.file.reader.writer.factory=com.pinterest.secor.io.impl.AvroParquetFileReaderWriterFactory -Dkafka.useTimestamp=true' \
    -e SECOR_KAFKA_TOPIC_FILTER='^avro_topics.*$' \
    -e SECOR_SCHEMA_REGISTRY=http://sr:8081 \
    -e SECOR_MESSAGE_PARSER='com.pinterest.secor.parser.AvroMessageParser' \
    -e S3_PATH='secor_parquet' \
    -e SECOR_MAX_FILE_SECONDS='300' \
    -e SECOR_MAX_FILE_BYTES='200000' \
    docker.repo.local/secor:0.27
````

The topic `avro_topics_x` in Kafka has Avro messages. As usual, the schema registered under the subject of `avro_topics_x-value`. Once the application started, it will throw the following exception when writing the messages:
```
Caused by: java.lang.IllegalStateException: Avro schema not found for topic avro_topics_x
        at com.pinterest.secor.common.SecorSchemaRegistryClient.getSchema(SecorSchemaRegistryClient.java:50)
        at com.pinterest.secor.io.impl.AvroParquetFileReaderWriterFactory$AvroParquetFileWriter.<init>(AvroParquetFileReaderWriterFactory.java:113)
        at com.pinterest.secor.io.impl.AvroParquetFileReaderWriterFactory.BuildFileWriter(AvroParquetFileReaderWriterFactory.java:56)
        at com.pinterest.secor.util.ReflectionUtil.createFileWriter(ReflectionUtil.java:154)
[...]

PR #951, opened 2019-09-22
Query Avro schema from schema registry if it is missing

irajhedayati:
As mentioned https://github.com/pinterest/secor/issues/894, secor fails if

- `secor.file.reader.writer.factory=com.pinterest.secor.io.impl.AvroParquetFileReaderWriterFactory` 
- `kafka.useTimestamp=true`

The writer is going to ask `SecorSchemaRegistryClient` for Avro schema for the writer. But that only has the schema in the cache if there has been a call for `decodeMessage`. However, if we use the message timestamp, it won't need to decode the message.

I updated the code to query the schema registry in order to update the cache. It uses the standard subject name for a Kafka topic as "kafka_topic-value" to query the schema.

### DIFF
--- a/src/main/java/com/pinterest/secor/common/SecorSchemaRegistryClient.java
+++ b/src/main/java/com/pinterest/secor/common/SecorSchemaRegistryClient.java
@@ -19,13 +19,16 @@
 package com.pinterest.secor.common;
 
 import io.confluent.kafka.schemaregistry.client.CachedSchemaRegistryClient;
+import io.confluent.kafka.schemaregistry.client.SchemaMetadata;
 import io.confluent.kafka.schemaregistry.client.SchemaRegistryClient;
+import io.confluent.kafka.schemaregistry.client.rest.exceptions.RestClientException;
 import io.confluent.kafka.serializers.KafkaAvroDeserializer;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericRecord;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.IOException;
 import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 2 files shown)


src/main/java/com/pinterest/secor/common/SecorSchemaRegistryClient.java
```diff
@@ -67,10 +70,27 @@ public GenericRecord deserialize(String topic, byte[] message) {
...
+     * @param topic a Kafka topic to query the schema for
+     * @return Schema object for the topic
+     * @throws IllegalStateException if there is no schema registered for this topic or it is not able to fetch it
+     */
     public Schema getSchema(String topic) {
         Schema schema = schemas.get(topic);
         if (schema == null) {
-            throw new IllegalStateException("Avro schema not found for topic " + topic);
+            try {
+                SchemaMetadata schemaMetadata = schemaRegistryClient.getLatestSchemaMetadata(topic + "-value");
```
HenryCaiHaiying: This seems odd to me.  Who defines that convention: 'topic_name-value'?

irajhedayati: basically, there are three strategies configurable in Kafka to work with Schema Registry explained [here](https://docs.confluent.io/current/schema-registry/serializer-formatter.html#sr-avro-subject-name-strategy). The default one which most of the deployments use is "TopicNameStrategy". In this strategy

- The subject name for the value is "topic-value"
- The subject name for the key is "key-value"

HenryCaiHaiying: is getSchema() only interested in value schema (or key schema as well)?

irajhedayati: Only value schema. Key is only an internal Kafka thing and is not gonna be stored in the Avro/AvroParquet files

HenryCaiHaiying:
Looks good.

PR #965, opened 2019-09-26
Test case for JsonORCFileReaderWriterFactory

suminb:
This PR adds a basic test case (file read/write roundtrip) for `JsonORCFileReaderWriterFactory`. The basic idea is mostly identical to that of `SequenceFileReaderWriterFactoryTest`.


HenryCaiHaiying:
Thanks for the contribution.

PR #983, opened 2019-09-30
Produce fat JAR (with all dependencies included)

suminb:
This modifies the `mvn package` command to produce a fat (uber) `.jar` file that includes all dependencies, so that one can simply drop a single `.jar` file to run secor. `mvn package` (or `make build`) command will make `secor-${VERSION}-SNAPSHOT-jar-with-dependencies.jar` under `targets` directory.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -443,6 +443,10 @@
                             <descriptors>
                                 <descriptor>src/main/assembly/secor.xml</descriptor>
                             </descriptors>
+                            <!-- get all project dependencies -->
+                            <descriptorRefs>
+                                <descriptorRef>jar-with-dependencies</descriptorRef>
+                            </descriptorRefs>
                         </configuration>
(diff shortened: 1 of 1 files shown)


HenryCaiHaiying:
Not all the people want a fat jar, if you want a fat jar it's better to create a different build target.  Even without the fat jar, you can use java -cp lib/* to depend on all the jars under lib/ folder.

suminb:
Agreed. Creating a different build target would make more sense. I'll probably work on this later.

As far as the `.gz` file produced by the `make build` command, I find it a bit easier to work with a single `.jar` file in my current situation :-)

suminb:
I thought about adding another lifecycle (e.g., `package-fatjar`), it doesn't seems like I can easily add one in Maven (`pom.xml`). I'll close this PR until I figure something out.

PR #1004, opened 2019-10-14
Reject non-string keys

suminb:
As far as I understand, JSON only allows string keys. However, the current implementation of `MapColumnConverter` allows arbitrary types as map keys which potentially cause issues for `JsonConverter`. This PR adds a safeguard that rejects non-string map keys.

### DIFF
--- a/src/main/java/com/pinterest/secor/util/orc/VectorColumnFiller.java
+++ b/src/main/java/com/pinterest/secor/util/orc/VectorColumnFiller.java
@@ -184,7 +184,7 @@ public class VectorColumnFiller {
         private void assertKeyType(TypeDescription schema) {
             TypeDescription keyType = schema.getChildren().get(0);
             String keyTypeName = keyType.getCategory().getName();
-            if (!keyTypeName.equals("string")) {
+            if (!keyTypeName.equalsIgnoreCase("string")) {
                 throw new IllegalArgumentException(
                         String.format("Unsupported key type: %s", keyTypeName));
             }
--- a/src/main/java/com/pinterest/secor/util/orc/VectorColumnFiller.java
+++ b/src/main/java/com/pinterest/secor/util/orc/VectorColumnFiller.java
@@ -41,7 +41,6 @@ import org.slf4j.LoggerFactory;
 import java.nio.charset.StandardCharsets;
 import java.sql.Timestamp;
 import java.util.List;
-import java.util.Set;
 
 /**
  * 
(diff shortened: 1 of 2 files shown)


src/main/java/com/pinterest/secor/util/orc/VectorColumnFiller.java
```diff
@@ -166,13 +165,28 @@ public void convert(JsonElement value, ColumnVector vect, int row) {
...
+                childConverters[c] = createConverter(childTypes.get(c));
+            }
+        }
+
+        /**
+         * Rejects non-string keys. This is a limitation imposed by JSON specifications that only allows strings
+         * as keys.
+         */
+        private void assertKeyType(TypeDescription schema) {
+            TypeDescription keyType = schema.getChildren().get(0);
```
HenryCaiHaiying: Does it guarantee that schema has Children?  And what are each children schema representing?

src/main/java/com/pinterest/secor/util/orc/VectorColumnFiller.java
```diff
@@ -166,13 +165,28 @@ public void convert(JsonElement value, ColumnVector vect, int row) {
...
+            String keyTypeName = keyType.getCategory().getName();
+            if (!keyTypeName.equals("string")) {
```
HenryCaiHaiying: Should it be case insensitive comparison?

src/main/java/com/pinterest/secor/util/orc/VectorColumnFiller.java
suminb: Suppose we have an ORC schema like `map<string,int>` then the first child is a `TypeDescription` instance representing `string` type. Not sure if `schema` is guaranteed to have at least one child. We might want to test with malformed ORC schema like `map<>`. It may or may not reach this point when such schema is given.

src/main/java/com/pinterest/secor/util/orc/VectorColumnFiller.java
suminb: That’s a good point. Not sure if ORC schema is case sensitive. But it would be a good idea to perform case insensitive comparison here just to be safe.

suminb:
Thanks for the review. I’ll come up with a better way to handle these concerns mentioned above and revise code accordingly.

src/main/java/com/pinterest/secor/util/orc/VectorColumnFiller.java
HenryCaiHaiying: Also make sure the passed-in 'schema' argument really represents the map schema

suminb: > Also make sure the passed-in 'schema' argument really represents the map schema

It certainly wouldn't hurt to do so, but I feel like this is somewhat redundant as it is already ensured by `VectorColumnFilter.createConverter()`. Its very logic doesn't leave any room for other types to fall into `MapColumnConverter`.

```
    public static JsonConverter createConverter(TypeDescription schema) {
        switch (schema.getCategory()) {

        // ...

        case MAP:
            return new MapColumnConverter(schema);
        default:
            throw new IllegalArgumentException("Unhandled type " + schema);
        }
    }
```

If you still insist doing so, I would suggest to add some common type checking logic in `JsonConverter` interface or in `VectorColumnFiller` class so that other types could also benefit from it. Any thoughts on this?

HenryCaiHaiying:
lgtm

PR #1011, opened 2019-10-16
WIP: Support union types

suminb:
While workin on #982, I've gained some knowledge on how ORC format works. So I might as well implement `UnionColumnConverter` to support union types.

This is more like a proposal/brainstorming rather than a complete solution. Any feedback and discussions are welcomed.


src/main/java/com/pinterest/secor/util/orc/VectorColumnFiller.java
```diff
@@ -200,6 +202,119 @@ public void convert(JsonElement value, ColumnVector vect, int row) {
...
+                default:
+                    throw new UnsupportedOperationException();
+            }
+        }
+
+        private byte getTypeMask(JsonPrimitive value) {
+            byte mask = 0;
+
+            mask |= value.isBoolean() ? BOOLEAN_MASK : 0;
+            mask |= value.isNumber()  ? NUMBER_MASK  : 0;
```
HenryCaiHaiying: Why continuously |=?  You can only has one type mask

HenryCaiHaiying:
lgtm

PR #1012, opened 2019-10-16
Avoid using wildcard import statements

suminb:
As for https://github.com/pinterest/secor/pull/982#discussion_r329724720, I thought it would be great if we could take a brief moment to re-organize some of the import statements. It may look like a big change, but there is no change functionality-wise.


HenryCaiHaiying:
lgtm

Issue #1010, opened 2019-10-16
Kafka offset checking logic in message write might lead to message loss 


breadpowder:
**Problem:**

Currently, a message writer enforces "offset increment by 1" checking before appending a msg into a local file. Otherwise, message writer deletes the underlying writer and existing local files per topic/partition.

The logic is in lines 82, see method adjustOffset in MessageWriter.class
  **_if (message.getOffset() != lastSeenOffset + 1) {
           ....
            mFileRegistry.deleteTopicPartition(topicPartition);
}_**

However, such offset increment "by 1" assumption is incorrect, see below stack overflow comments regarding both transactional and non-transactional mode (our company also observes the same behaviour in production, transactional mode is even worse, since the offset in the subsequent transaction is indeed NOT decremental by 1). Therefore, such check causes local files be deleted hence message loss. 
https://stackoverflow.com/questions/54636524/kafka-streams-does-not-increment-offset-by-1-when-producing-to-topic

**Fix:**

We need to refactor this offset checking logic:
The logic is to guard against a potential rebalance in order to remove duplicates.  However, to guard such condition we only need to ensure the offset is incremental. Since in Kafka,  rebalance or fault tolerance mechanism will "replay" messages by reassigning partition to a consumer from the last committed offset, a consumer won't loss message. Therefore, We only need to ensure current message(processing) offset is no less than last seen offset (processed). If not, we trim offsets between processing offset (exclusive) and last seen offset (processed) .

**Next:**
Will push a PR regarding the fix. 

Please comment. Thank you!

HenryCaiHaiying:
Based on my observation, offset is incremented by 1 in non-transactional
mode.

The code above is to guard against the situation when consumer rebalance
happens, the other consumer already uploaded the specified offsets to S3 so
the current consumer no longer needs to hold and process those offsets.

breadpowder:
A problem is the logic does not hold for transactional mode, where holds no guarantee of offset +1.

My question is for non-transaction mode: let's assumes that the offset is incremented by 1 assumption holds for now , how could a processing message offset > lastSeenOffset +1 happens?  I am trying to understand this so we can simplify the logic for both modes.

Issue #682, opened 2019-04-22
Unable To Process Transaction API Events 


kautshukla:
I'm using kafak_2.0.0 profile with https://github.com/pinterest/secor/blob/master/src/main/config/kafka-2.0.0/kafka.properties properties to process the events which are pushed through Kafka producer using transaction API. While writing to an intermediate log file only captures the last event rest get skipped, found the same in the uploaded file as well.

Why it is happening? 

Code base: 0.28-SNAPSHOT

I found the bug where the offset change isn't by +1 in case of read_commited isolation for Kafka consumer, whereas consuming the Kafka transaction API events pushed.

Block : https://github.com/pinterest/secor/blob/4068d9d3f6fa2bc2d30ad352973e044df4ccf4db/src/main/java/com/pinterest/secor/writer/MessageWriter.java#L78

There should be the check if new consumer which consumes {Kafka producer using transaction API} produced events which have offset difference of +2 as well.

kautshukla:
@HenryCaiHaiying could you please have a look in this issue.

HenryCaiHaiying:
Duplicate of https://github.com/pinterest/secor/pull/1017

PR #1017, opened 2019-10-18
Message offset adjust ticket #1010

breadpowder:
This is for ticket id #1010 regarding offset.

The offset validation logic is refactored a bit to ensure working for both transactional and non-transactional mode. Since:
1) in transactional mode messages are not +1 between messages
2) in non-transaction mode, there might be very minor chances a message get delivered more than once due to retrying.

Now the offset logic in message writer works:  If the current processing message offset is not greater than "last seen offset" (messages have written), it means there are rebalances and/or the broker is replaying message from last committed offset/failure point. Since the messages before offsets are written, we can safely trim those message to remove duplicates. 

The existing commit offset validation logic in upload class remains intact . The existing logic ensures that after rebalances, if another consumer has made progress on a same topic/partition, we either trims those files or trim corresponding completed offset.

### DIFF
--- a/src/main/java/com/pinterest/secor/common/OffsetTracker.java
+++ b/src/main/java/com/pinterest/secor/common/OffsetTracker.java
@@ -36,6 +36,10 @@ public class OffsetTracker {
     private HashMap<TopicPartition, Long> mCommittedOffsetCount;
 
     public OffsetTracker() {
+        reInitiateOffset();
+    }
+
+    public void reInitiateOffset() {
         mLastSeenOffset = new HashMap<TopicPartition, Long>();
         mCommittedOffsetCount = new HashMap<TopicPartition, Long>();
         mFirstSeendOffset = new HashMap<TopicPartition, Long>();
@@ -52,15 +56,14 @@ public class OffsetTracker {
     public long setLastSeenOffset(TopicPartition topicPartition, long offset) {
         long lastSeenOffset = getLastSeenOffset(topicPartition);
         mLastSeenOffset.put(topicPartition, offset);
-        if (lastSeenOffset + 1 != offset) {
-            if (lastSeenOffset >= 0) {
-                LOG.warn("offset for topic {} partition {} changed from {} to {}",
-                        topicPartition.getTopic(),topicPartition.getPartition(),lastSeenOffset, offset);
-            } else {
+        if (offset < lastSeenOffset + 1) {
+            LOG.warn("offset for topic {} partition {} goes back from {} to {}",
+                    topicPartition.getTopic(), topicPartition.getPartition(), lastSeenOffset, offset);
+
+        } else if (lastSeenOffset < 0) {
                 LOG.info("starting to consume topic {} partition {} from offset {}",
                         topicPartition.getTopic(),topicPartition.getPartition(),offset);
             }
-        }
         if (mFirstSeendOffset.get(topicPartition) == null) {
             mFirstSeendOffset.put(topicPartition, offset);
         }
--- a/src/main/java/com/pinterest/secor/writer/MessageWriter.java
+++ b/src/main/java/com/pinterest/secor/writer/MessageWriter.java
@@ -75,14 +75,15 @@ public class MessageWriter {
         mGeneration = mConfig.getGeneration();
     }
 
-    public void adjustOffset(Message message) throws IOException {
+    public void adjustOffset(Message message, boolean isLegacyConsumer) throws IOException {
         TopicPartition topicPartition = new TopicPartition(message.getTopic(),
                                                            message.getKafkaPartition());
         long lastSeenOffset = mOffsetTracker.getLastSeenOffset(topicPartition);
-        if (message.getOffset() != lastSeenOffset + 1) {
+        //this validation logic is for duplicates removing as no rebalancing callbacks is incompatible with LegacyKafkaMessageIterator
+        if (isLegacyConsumer && message.getOffset() != lastSeenOffset + 1) {
             StatsUtil.incr("secor.consumer_rebalance_count." + topicPartition.getTopic());
             // There was a rebalancing event since we read the last message.
-            LOG.debug("offset of message {} does not follow sequentially the last seen offset {}. " +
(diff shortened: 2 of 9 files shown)


HenryCaiHaiying:
There are too many files changed in this PR.  Please check your coding style.  We don't want to use wildcard imports, each individual package needs to be explicitly specified and imported.

breadpowder:
Done. Please use the latest one. This one is rebased on most recent master and leaves my commits clear to read.

breadpowder:
How about a bit enhancement over the above approach?  Instead of removing the underlying files when rebalancing happens, let's upload the file and commit the offset in the partition revoked notification. Therefore, other consumer don't need to rewrite the same data.

HenryCaiHaiying:
If the rebalance happens, the files the current consumer was working on are no longer source of truth.  The other consumer can already upload all or part of those offsets to S3.  The other situation is the consumer was working on partition 0 for some time, then rebalance event 1 happens, he lost partition 0 but he still keep some unfinished partition 0 in local filesystem.  After sometime another rebalance event 2 happens, he got back partition 0 but the offset jumps ahead already.  If he continues appending the new messages from partition 0 to local files, those files will have a gap on message offsets.  So it's best to clear all local files when offset inconsistency discovers.

The messages are  not lost because some other consumers already uploaded those offsets to S3.

The transactional message is a bit different story, that's why we need to distinguish between this case and consumer rebalance.

breadpowder:
Thank you! Here, I understand and agree with your points here regarding the other customer took over and the possibility about offset jump when a previous partition assigned back .

Here, my thinking is we don't need to clear the local files to ensure consistency, instead, we can do a upload and commit offset in the "partition revoke" callback given the kafka consumerRebalanceListener Api's promises (http://kafka.apache.org/20/javadoc/index.html?org/apache/kafka/clients/consumer/ConsumerRebalanceListener.html). The api document says "this method will be called before a rebalance operation starts and after the consumer stops fetching data. It is recommended that offsets should be committed in this callback to either Kafka or a custom offset store to prevent duplicate data". 

Therefore, we can safely upload and commit offset after process existing messages buffer in the "partition revoke callback". It won't cause data inconsistency issue given the other consumer will start from the newly committed offset, as they work sequentially without any offset overlapping. 

By doing this, the benefit is that the other customer don't need to reprocess the same data. Let me know your thoughts.   Thank you again for sharing your thoughts!

HenryCaiHaiying:
If you add the uploading code in the ConsumerRebalanceListener, the
uploading will clear the local files automatically so the old check and
file clear might not be really needed.  It's probably also a good idea to
clear the lastSeen data structure on that rebalance listener so everything
start fresh after rebalance.

We don't have that consumer rebalance listener before, but this new
listener we actually can do things more cleanly.

breadpowder:
Sure. will commit soon.

HenryCaiHaiying:
Looking forward to your PR

breadpowder:
As discussed, rebalance logic is added into SecorConsumerRebalanceListener and RebalanceHandler.  Please review.

src/main/java/com/pinterest/secor/common/OffsetTracker.java
```diff
@@ -52,15 +52,15 @@ public long getLastSeenOffset(TopicPartition topicPartition) {
     public long setLastSeenOffset(TopicPartition topicPartition, long offset) {
         long lastSeenOffset = getLastSeenOffset(topicPartition);
         mLastSeenOffset.put(topicPartition, offset);
-        if (lastSeenOffset + 1 != offset) {
-            if (lastSeenOffset >= 0) {
-                LOG.warn("offset for topic {} partition {} changed from {} to {}",
-                        topicPartition.getTopic(),topicPartition.getPartition(),lastSeenOffset, offset);
-            } else {
+        if (offset < lastSeenOffset + 1) {
+            LOG.warn("offset for topic {} partition {} goes back from {} to {}",
```
HenryCaiHaiying: How about we throw an exception here?  I don't think we should see this happen with the code change in Rebalance listener?

src/main/java/com/pinterest/secor/writer/MessageWriter.java
```diff
@@ -75,14 +75,15 @@ public MessageWriter(SecorConfig config, OffsetTracker offsetTracker,
...
     }
 
-    public void adjustOffset(Message message) throws IOException {
+    public void adjustOffset(Message message, boolean isLegacyConsumer) throws IOException {
         TopicPartition topicPartition = new TopicPartition(message.getTopic(),
                                                            message.getKafkaPartition());
         long lastSeenOffset = mOffsetTracker.getLastSeenOffset(topicPartition);
-        if (message.getOffset() != lastSeenOffset + 1) {
+        //this validation logic is for duplicates removing as no rebalancing callbacks is incompatible with LegacyKafkaMessageIterator
+        if (isLegacyConsumer && message.getOffset() < lastSeenOffset + 1) {
```
HenryCaiHaiying: Should you use the old logic: 'message.getOffset() != lastSeenOffset +1' here?

src/main/java/com/pinterest/secor/common/OffsetTracker.java
breadpowder: For compatibility with LegacyKafkaMessageIterator, no such listener exists, so we have to keep this for now.

src/main/java/com/pinterest/secor/writer/MessageWriter.java
breadpowder: sure. will do an update.

src/main/java/com/pinterest/secor/common/OffsetTracker.java
breadpowder: Would prefer to use a separator class for RebalanceListener since rebalance logic is separator concern from message iterator including upload and cleanup.

breadpowder: Let me know if the above is ok, will push shortly.

breadpowder:
Just happened to find that this issue linked another ticket #682

HenryCaiHaiying:
lgtm, thanks for the effort.

PR #1018, opened 2019-10-20
Follow-up of #1011 (Support union types)

suminb:
#1011 was missing a critical part: writing union type rows as JSON objects. `FileWriter` works okay with union types, but when we try to read values via `FileReader` it sees incomplete JSON data.

To illustrate the problem a bit further, `FileWriter` doesn't quite finish writing ORC data as JSON because `JsonFieldFiller` is not fully implemented to cope with union types.

```
            case UNION:
                // printUnion(writer, (UnionColumnVector) vector, schema, row);
                break;
```

And thus `FileReader` would encounter incomplete JSON values like:

```
{"values":}
```

This PR addresses the issue.


HenryCaiHaiying:
lgtm

PR #1108, opened 2019-12-10
Only perform Travis builds on the 'master' branch

jparise:
Pull Requests will continue to be built, as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+branches:
+  only:
+    - "master"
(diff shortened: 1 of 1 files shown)


HenryCaiHaiying:
What was the purpose of the PR?  I thought secor only has one branch: master

jparise:
It was kicking off two builds when branch was pushed to the main repository. Dependabot does this, and the second build was redundant because it was already doing a PR build.

Issue #455, opened 2018-09-11
No included message parser that gets timestamp from record metadata?


kennydo:
Hi! I am trying to group my S3 output keys by hour and minute based on the timestamp in each Kafka record's metadata (the output of `timestamp ()` from https://kafka.apache.org/0100/javadoc/index.html?org/apache/kafka/connect/data/Timestamp.html). 

The `TimestampedMessageParser` class gets me 99.9% of the way there, to the point that all I needed to do was extend it in a super small class. I've got a branch with this ready and it's trivially simple: https://github.com/pinterest/secor/compare/master...kennydo:kedo-kafka-record-timestamp-parser?expand=1

Is there a reason the functionality (of using the timestamp in each record's metadata) isn't part of the standard set of message parsers? Could/should this simple extension of `TimestampedMessageParser` be merged upstream?

HenryCaiHaiying:
I think what you asked for is already there.  Take a look at Kafka10MessageTimestamp.java

kennydo:
`Kafka10MessageTimestamp` extracts the timestamp out of the Kafka record (and my secor is already configured to use it), but what I'm more interested in is making use of that timestamp data to partition partition the records into the correct hour/minute bucket, which I believe i the role of the subclasses of `MessageParser` (and `TimestampedMessageParser`).

kennydo:
Yea, I've enabled that config parameter.

I then found out that `TimestampedMessageParser` is an abstract class, with an abstract function that is not implemented (https://github.com/pinterest/secor/blob/master/src/main/java/com/pinterest/secor/parser/TimestampedMessageParser.java#L150), so Secor is unable to instantiate it at all.

Looking closer, it looks like the actual logic in my branch isn't actually necessary. I just needed a concrete class that implements `extractTimestampMillis` to appease the compiler.

So this issue boils down to: is there a reason `TimestampedMessageParser` is abstract? Could `extractTimestampMillis` be moved elsewhere so that `TimestampedMessageParser` doesn't have to be abstract?

HenryCaiHaiying:
The code should go to call toMills() if you already set mUseKafkaTimestamp, you are looking at the wrong path.

(mUseKafkaTimestamp) ? toMillis(message.getTimestamp())

kennydo:
Right, the code path that leads to calling `extractTimestampMillis` is never called because I've configured it like so.

[...]

PR #1147, opened 2020-01-13
Use timestamp from Kafka message header

suminb:
Although `JsonMessageParser` provides a neat way to extract a timestamp column from JSON messages, my team faced a unique set of challenges. To give you a little bit of background, we are consuming about 20-30 Kafka topics with a single *deployment* (multiple Secor instances,  identical configs). We would like to load data to S3 by partitioning records by date, and we would like to use the timestamp in Kafka message headers for this purpose.

I understand `TimestampedMessageParser`'s `getTimestampMills()` could extract the Kafka message timestamp as follows:

```
    public long getTimestampMillis(Message message) throws Exception {
        return (mUseKafkaTimestamp) ? toMillis(message.getTimestamp()) : extractTimestampMillis(message);
    }
```

However, `JsonMessageParser` overwrites it with its own implementation, and there is no way to retrieve the value of `message.getTimestamp()`.

```
    @Override
    public long extractTimestampMillis(final Message message) {
        JSONObject jsonObject = (JSONObject) JSONValue.parse(message.getPayload());
        if (jsonObject != null) {
            Object fieldValue = getJsonFieldValue(jsonObject);
            if (fieldValue != null) {
                return toMillis(Double.valueOf(fieldValue.toString()).longValue());
            }
        } else if (m_timestampRequired) {
            throw new RuntimeException("Missing timestamp field for message: " + message);
        }
        return 0;
    }
```

I've also looked at https://github.com/pinterest/secor/issues/455, but it does not seems like `kafka.message.timestamp.className` has any effect in `JsonMessageParser`.

Hence, I would like to propose this as a quick solution. If you think we'd better off fixing `JsonMessageParser.extractTimestampMillis()` somehow, please let me know. Thanks.

### DIFF
--- a/src/main/java/com/pinterest/secor/parser/KafkaTimestampParser.java
+++ b/src/main/java/com/pinterest/secor/parser/KafkaTimestampParser.java
@@ -1,0 +1,15 @@
+package com.pinterest.secor.parser;
+
+import com.pinterest.secor.common.SecorConfig;
+import com.pinterest.secor.message.Message;
+
+public class KafkaTimestampParser extends TimestampedMessageParser {
+    public KafkaTimestampParser(SecorConfig config) {
+        super(config);
+    }
+
+    @Override
+    public long extractTimestampMillis(Message message) {
+        return message.getTimestamp();
+    }
+}


HenryCaiHaiying:
From this line: 
        return (mUseKafkaTimestamp) ? toMillis(message.getTimestamp()) : extractTimestampMillis(message);

if you have mUseKafkaTimestap, it will not go to extractTimestampMillis()

suminb:
I think here is what happened. I initially put `kafka.useTimestamp=true` in `secor.prod.partition.properties` and observed that Secor was still trying to extract timestamp value from JSON message bodies. Here, `secor.prod.partition.properties` includes ` secor.prod.properties` which includes `secor.common.properties` which then includes `kafka.properties` that defines `kafka.useTimestamp=false`.

After modifying the `kafka.properties` file, everything seems working as intended (with `secor.message.parser.class=com.pinterest.secor.parser.JsonMessageParser`).

I'm closing this PR as the issue I had can be resolved without any code modification. Thanks for your feedback.

Issue #1251, opened 2020-04-01
ProgressMonitor fails when one partition fails


f1yegor:
For the monitoring purposes, I think ProgressMonitor shouldn't fail when one partition stat is failing:
```
(com.pinterest.secor.main.ProgressMonitorMain$1:52) ERROR Progress monitor failed
java.lang.RuntimeException: unable to fetch message after 10 Retries
at com.pinterest.secor.common.SecorKafkaClient.readSingleMessage(SecorKafkaClient.java:98)
at com.pinterest.secor.common.SecorKafkaClient.getCommittedMessage(SecorKafkaClient.java:78)
at com.pinterest.secor.tools.ProgressMonitor.getStats(ProgressMonitor.java:208)
at com.pinterest.secor.tools.ProgressMonitor.exportStats(ProgressMonitor.java:144)
```

HenryCaiHaiying:
What was the reason the fetch failed?  The message no longer existed?

f1yegor:
yes, we have some topics that get deprecated and have no messages anymore

HenryCaiHaiying:
In the case, I feel it might better to remove those topics from secor config and restart secor.  If we modify the code to ignore those errors for ProgressMonitor it might hide the real errors for other users.

f1yegor:
1. doesn't it fail when the message is absent after `mPollTimeout` interval? 
my topic was with rare messages published by old client from time to time. 
2. it's still `LOG.warn` to be visible

HenryCaiHaiying:
I see your point, it's actually pretty easy to fix.  In
SecorKafkaClient.java:

if (message == null) {
    throw new RuntimeException("unable to fetch message after " +
MAX_READ_POLL_ATTEMPTS + " Retries");
}
return message;

You can replace the exception thrown with a LOG.warn()

Do you want to submit a PR for this?  I can merge it in quickly.

f1yegor:
https://github.com/pinterest/secor/pull/1346

HenryCaiHaiying:
Left a comment in the PR

HenryCaiHaiying:
Merged in the PR.

PR #1252, opened 2020-04-01
ProgressMonitor: Handle failure of partition

f1yegor:
Addresses https://github.com/pinterest/secor/issues/1251


HenryCaiHaiying:
Thanks for the contribution.

PR #1269, opened 2020-04-10
docker-entrypoint.sh shouldn't use Bash "if" syntax

marcin-kamionowski:
I tried to use Secor with Docker, but there were problems with syntax of docker-entrypoint.sh script. 
As a result, env variables were not used to generate Secor config.
This patch fix replaces Bash-like syntax with POSIX one.


HenryCaiHaiying:
Looks good to me.  We haven't used docker so it's hard to verify those scripts.

PR #1309, opened 2020-05-06
JSON array of object saved as ORC format

marcin-kamionowski:
Hi
In some situations Kafka messages may contains JSON array of objects. Each item on the list have the same schema. This patch allows to read messages like this and save each list item as a separate records.
This patch should not have performance implications as root JSON element type was checked before.
It was expected to receive JsonObject but in our scenario it is JsonArray.

### DIFF
--- a/src/main/java/com/pinterest/secor/io/impl/JsonORCFileReaderWriterFactory.java
+++ b/src/main/java/com/pinterest/secor/io/impl/JsonORCFileReaderWriterFactory.java
@@ -20,15 +20,16 @@ package com.pinterest.secor.io.impl;
 
 import java.io.IOException;
 import java.io.StringWriter;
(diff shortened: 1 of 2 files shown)


HenryCaiHaiying:
It looks like the travis build is failing for this PR: https://travis-ci.org/github/pinterest/secor/builds/683900104

The error doesn't seem related to your PR, and probably a transient error.

Can you do a dummy checkin (e.g adding a comment in the code) to trigger the rebuild?

marcin-kamionowski:
Now it passed: https://travis-ci.org/github/pinterest/secor/builds/684173873

HenryCaiHaiying:
Great, I will merge the PR, thanks.

marcin-kamionowski:
No problem, thanks!

PR #1325, opened 2020-05-18
Add more metrics for uploader:

HenryCaiHaiying:
- uploader.offset_mismatches: count how many times upload fails;

- uploader.partition_trims: count how many times partition trimming happens;
        // There was a rebalancing event and someone committed an offset lower than that of the current message.  We need to trim local files.
- uploader.partition_deletes: count how many times partition deletion happens;
        // There was a rebalancing event and someone committed an offset beyond that of the current message.  We need to delete the local file.
	// zookeeper committed offset didn't match for topic/partition


HenryCaiHaiying:
@gmcoringa Added a few more metrics per your request

HenryCaiHaiying:
Will do the micrometer integration in the next PR

PR #1346, opened 2020-05-24
ProgressMonitor: Handle failure of partition

f1yegor:
Modification of previous https://github.com/pinterest/secor/pull/1252 after discussion with @HenryCaiHaiying


HenryCaiHaiying:
The change on SecorKasfkaClient.java looks fine but I am not sure whether we should remove the RuntimeException handling in ProgressMonitor.java, there might be other places which can still throw NPE.

f1yegor:
Corrected

HenryCaiHaiying:

PR #1408, opened 2020-06-24
Add max.poll.interval.ms kafka producer setting

jehuty0shift:
This PR has for goal to enable the configuration of max.poll.interval.ms. As stated in the documentation (https://kafka.apache.org/documentation/#max.poll.interval.ms). This setting can prevent rebalancing during any long upload in Secor.


HenryCaiHaiying:

Issue #785, opened 2019-06-18
Secor is leaking memory through DeleteOnExit


giner:
Secor is exhausting the while heap over time and either crashes or starts spinning due to GC trying to free memory very frequently. Memory is accumulated in the **DeleteOnExitHook** by **DeleteOnExit** method. We are running Secor v0.25 and the related code seems to be the same in the latest version (which is v0.27 at the moment of writing).

![secor - memory usage over time](https://user-images.githubusercontent.com/533434/59649270-02015900-91bd-11e9-8a65-286b3c73add5.png)
![secor - memory dump analysis 1](https://user-images.githubusercontent.com/533434/59649276-07f73a00-91bd-11e9-8e51-d6a2a04d9b82.png)
![secor - memory dump analysis 2](https://user-images.githubusercontent.com/533434/59649280-0f1e4800-91bd-11e9-8f42-cf812c7002cb.png)

dovka:
We are getting the same problem - every few days secor crashes with identical Memory and CPu patterns. Filed under #964

yehoshoua:
any updates ?

HenryCaiHaiying:
This is interesting.  If you can reproduce the problem, can you turn on the debug logging category for FileRegistry.java which will prints out all the files added for deleteOnExit.

It would be interesting to see why there will be so many files added, we have never seen the memory problem on DeleteOnExitHook before and DeleteOnExitHook doesn't have a method to clear the map either.  One suspicion I have is the file size is set to too small (secor.max.file.size.bytes and secor.max.file.age.seconds) so it ends up creating too many files.

pdambrauskas:
I think I have the same problem. 
In my case, I have custom `MessageParser` implementation, which adds multi-level partitions, something like: `/owner=1/workspace=2/event=click/dt=2020-07-07` in that case, we can get really big number of files and storing them in compressed parquet ads up to even biger memory consumption.

Any ideas on how to approach this problem?
Maybe we should pause Kafka consumers, when we reach X number of files, upload those files and resume? But in that case we'll potentially have many small files :thinking:.  We could also partition kafka topic by same values as we partition our data on  destination system, in that case consumed data would appear on same consumer thread and that would help us to have smaller number of files?

HenryCaiHaiying:
Looks good to me, I left a small comment in the code for backward compatible behavior.   And I don't think we should sort the files since that would change the file uploading order, some people might not like that.

HenryCaiHaiying:
[...]

PR #1443, opened 2020-07-08
Add secor.max.active.files property to controll active file count

pdambrauskas:
Partially addresses https://github.com/pinterest/secor/issues/785

Keeps active file under `secor.max.file.count` (default 100)

@HenryCaiHaiying , what do you think, is this the right way to approach this problem? I was thinking, maybe Files should be sorted by size before calling `checkTopicPartition` so that the biggest file is uploaded first, but decided to keep that simple for now.

Tried it for our use case (big number of partitions in small amount of time) it helped with OutOfMemory exception.

### DIFF
--- a/src/main/java/com/pinterest/secor/common/SecorConfig.java
+++ b/src/main/java/com/pinterest/secor/common/SecorConfig.java
@@ -277,6 +277,10 @@ public class SecorConfig {
         return getLong("secor.max.file.age.seconds");
     }
 
+    public int getMaxActiveFiles() {
+        return getInt("secor.max.file.count", -1);
+    }
+
     public boolean getUploadOnShutdown() {
         return getBoolean("secor.upload.on.shutdown");
     }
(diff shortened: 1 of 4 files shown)


src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -277,6 +277,10 @@ public long getMaxFileAgeSeconds() {
         return getLong("secor.max.file.age.seconds");
     }
 
+    public int getMaxActiveFiles() {
+        return getInt("secor.max.file.count", 100);
```
HenryCaiHaiying, line 281: How about we set the default to -1, and on the caller side if getMaxActiveFiles returns -1 it will not forceUpload, this way we can maintain the backward compatible behavior for the installations who didn't face memory problem.

HenryCaiHaiying:
Looks good to me, I left a small comment in the code for backward compatible behavior.   And I don't think we should sort the files since that would change the file uploading order, some people might not like that.

pdambrauskas:
@HenryCaiHaiying updated. Changed default value to `-1` and made feature effective, only when value is > -1.

HenryCaiHaiying:

PR #1449, opened 2020-07-10
Rework cleanup process on shutdown

pdambrauskas:
fixes https://github.com/pinterest/secor/issues/785

Due to [this issue](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=6664633) `DeleteOnExit` hook cannot be used on long-running applications.

This PR replaces usage of DeleteOnExitHook with custom shut down hook, which cleans up local storage directory.

@giner @HenryCaiHaiying does this makes sense :) ?

### DIFF
--- a/src/main/java/com/pinterest/secor/main/ConsumerMain.java
+++ b/src/main/java/com/pinterest/secor/main/ConsumerMain.java
@@ -20,9 +20,12 @@ package com.pinterest.secor.main;
 
 import com.pinterest.secor.common.OstrichAdminService;
 import com.pinterest.secor.common.SecorConfig;
+import com.pinterest.secor.common.ShutdownHookRegistry;
 import com.pinterest.secor.consumer.Consumer;
+import com.pinterest.secor.io.StagingDirectoryCleaner;
 import com.pinterest.secor.tools.LogFileDeleter;
 import com.pinterest.secor.util.FileUtil;
+import com.pinterest.secor.util.IdUtil;
 import com.pinterest.secor.util.RateLimitUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -54,6 +57,8 @@ public class ConsumerMain {
         }
         try {
             SecorConfig config = SecorConfig.load();
+            String stagingDirectoryPath = config.getLocalPath() + '/' + IdUtil.getLocalMessageDir();
+            ShutdownHookRegistry.registerHook(10, new StagingDirectoryCleaner(stagingDirectoryPath));
             OstrichAdminService ostrichService = new OstrichAdminService(config.getOstrichPort());
             ostrichService.start();
             FileUtil.configure(config);
--- a/src/main/java/com/pinterest/secor/common/FileRegistry.java
+++ b/src/main/java/com/pinterest/secor/common/FileRegistry.java
@@ -139,11 +139,6 @@ public class FileRegistry {
             mWriters.put(path, writer);
             mCreationTimes.put(path, System.currentTimeMillis() / 1000L);
             LOG.debug("created writer for path {}", path.getLogFilePath());
-            LOG.debug("Register deleteOnExit for path {}", path.getLogFilePath());
-            FileUtil.deleteOnExit(path.getLogFileParentDir());
-            FileUtil.deleteOnExit(path.getLogFileDir());
-            FileUtil.deleteOnExit(path.getLogFilePath());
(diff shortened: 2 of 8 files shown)


src/main/java/com/pinterest/secor/main/ConsumerMain.java
```diff
@@ -54,6 +55,9 @@ public static void main(String[] args) {
         }
         try {
             SecorConfig config = SecorConfig.load();
+            if (!config.getUploadOnShutdown()) {
```
pdambrauskas, line 58: Only adding this hook, if uploadOnShutdown is disabled, otherwise we cannot be sure that deletion will happen after upload. Java does not provide strong order for shutdown hooks, they can be executed in any order & in parallel.

Other option would be implementing some kind of "shutdown hook registry" combining all hooks into one thread and registring all Runnables as one shutdown hook. But I think we do not need that, since files are deleted after upload anyways.

src/main/java/com/pinterest/secor/io/StagingCleaner.java
```diff
@@ -0,0 +1,53 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Runnable used to delete staging folder content.
+ * Deletes folders content, while keeping folder itself.
+ *
+ * @author Paulius Dambrauskas (p.dambrauskas@gmail.com)
+ *
+ */
+public class StagingCleaner implements Runnable {
```
HenryCaiHaiying, line 34: Can you rename the class name StagingCleaner to StagingDirectoryCleaner?

src/main/java/com/pinterest/secor/io/StagingCleaner.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private final File mStagingDir;
+
+    public StagingCleaner(String stagingPath) {
+        this.mStagingDir = new File(stagingPath);
+    }
+
+    @Override
+    public void run() {
+        for (File file : mStagingDir.listFiles()) {
```
HenryCaiHaiying, line 45: I think we can just call FileUtils.deleteDirectory(stagingPath) instead of using a loop.

src/main/java/com/pinterest/secor/main/ConsumerMain.java
```diff
@@ -54,6 +55,9 @@ public static void main(String[] args) {
...
+                Runtime.getRuntime().addShutdownHook(new Thread(new StagingCleaner(config.getLocalPath())));
```
HenryCaiHaiying, line 59: Please use this path instead (Take a look at Upload.java):  config.getLocalPath() is a shared folder with other processes/threads running on that host.

                        String localPrefix = mConfig.getLocalPath() + '/' +
                            IdUtil.getLocalMessageDir();

src/main/java/com/pinterest/secor/main/ConsumerMain.java
HenryCaiHaiying: Take a look at DeleteOnExitHook.java, there seems a way to register your hook as the last one.  If ordering cannot be done, I think you can modify the hook in uploadOnShutdown to have them call your hook's folder deletion methods.  It's OK to have both hooks to call delete folders twice.

HenryCaiHaiying:
Left some comments, it's a good fix.  I think we are using deleteOnExit() just in case the normal deletion didn't happen, by looking at FileRegistry.deletePath(), the local file would normally be deleted after upload, otherwise you would see millions of local files on disk during secor running (which would be a out of disk problem rather than out of memory problem).

pdambrauskas:
@HenryCaiHaiying thanks for review!

I've updated shutdown hook logics.

- Renamed delete hook file, as you suggested
- Changed deletion path
- Deleting whole path, instead of files inside
- Created Shutdown hook registry, to make it easier to run hooks in specified order.

I decided not tou use `DeleteOnExitHook` approch, because documentation of `registerShutdownHook` method states that it is used only for `DeleteOnExitHook`. I'm afraid using it somewhere else might break something. I tried it on my first attemt, and integration tests failed in some strange way: https://travis-ci.org/github/pinterest/secor/jobs/706801339
> Registers a shutdown hook.
   It is expected that this method with registerShutdownInProgress=true
   is only used to register DeleteOnExitHook since the first file
   may be added to the delete on exit list by the application shutdown
   hooks.

Also having single Shut down hook registry looks a bit nicer, as it makes usage of shutdown hooks more visible.

HenryCaiHaiying:
Looks good, this is a good fix.

Issue #430, opened 2018-06-10
Feature: Native prometheus metrics


norrs:
Provide native prometheus `/metrics` endpoint instead of converting statsd->prometheus metrics which is currently done in the Helm charts for kubernetes

https://prometheus.io/ supported by https://www.cncf.io/

HenryCaiHaiying:
With https://github.com/pinterest/secor/issues/1341, there is integration with micrometer.io which supports prometheus, it should be pretty easy to add prometheus through micrometer.

PR #1452, opened 2020-07-12
Add native support for Prometheus

pdambrauskas:
We are running Secor on K8s and monitoring our stack with Prometheus.
As Secor uses Micrometer, it is not that hard to add Prometheus support directly, and avoid using exporters for Secor metrics.

I've changed the way `gauge` method works, because Prometheus endpoint was returning `NaN` values.
More on that https://stackoverflow.com/questions/50821924/micrometer-prometheus-gauge-displays-nan

I've added `/prometheus` endpoint to Ostrich Server to avoid making things complicated with listening to some other port.

Also moved initialization of metric controller up to `ConsumerMain` class. I guess we don't need to have different instances for each Consumer. Otherwise  I'd have to ensure, that there is only one instance of PrometheusMetricRegistry to avoid duplications  and other problems on Prometheus http endpoint. This shouldnt be a problem, as Metric registries are delegating calls to static methods anyway.

ref: https://github.com/pinterest/secor/issues/430

### DIFF
--- a/src/main/java/com/pinterest/secor/common/monitoring/OstrichAdminService.java
+++ b/src/main/java/com/pinterest/secor/common/monitoring/OstrichAdminService.java
@@ -16,8 +16,9 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package com.pinterest.secor.common;
+package com.pinterest.secor.common.monitoring;
 
+import com.pinterest.secor.common.SecorConfig;
 import com.pinterest.secor.util.StatsUtil;
 import com.twitter.ostrich.admin.AdminServiceFactory;
 import com.twitter.ostrich.admin.CustomHttpHandler;
@@ -30,6 +31,7 @@ import scala.Option;
 import scala.collection.JavaConversions;
 import scala.collection.Map$;
 import scala.collection.immutable.List$;
+import scala.collection.immutable.Map;
 import scala.util.matching.Regex;
 
 import java.util.Arrays;
@@ -44,13 +46,17 @@ import java.util.concurrent.TimeUnit;
 public class OstrichAdminService {
     private static final Logger LOG = LoggerFactory.getLogger(OstrichAdminService.class);
     private final int mPort;
+    private final boolean mPrometheusEnabled;
 
-    public OstrichAdminService(int port) {
-        this.mPort = port;
+    public OstrichAdminService(SecorConfig config) {
+        mPort = config.getOstrichPort();
+        mPrometheusEnabled = config.getMicroMeterCollectorPrometheusEnabled();
     }
 
     public void start() {
         Duration[] defaultLatchIntervals = {Duration.apply(1, TimeUnit.MINUTES)};
+        Map<String, CustomHttpHandler> handlers = mPrometheusEnabled ?
+                new Map.Map1<>("/prometheus", new PrometheusHandler()) : Map$.MODULE$.empty();
         @SuppressWarnings("deprecation")
         AdminServiceFactory adminServiceFactory = new AdminServiceFactory(
             this.mPort,
@@ -58,7 +64,7 @@ public class OstrichAdminService {
             List$.MODULE$.<StatsFactory>empty(),
             Option.<String>empty(),
             List$.MODULE$.<Regex>empty(),
-            Map$.MODULE$.<String, CustomHttpHandler>empty(),
+            handlers,
             JavaConversions
                 .asScalaBuffer(Arrays.asList(defaultLatchIntervals)).toList()
         );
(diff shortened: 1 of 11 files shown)


src/main/java/com/pinterest/secor/common/monitoring/OstrichAdminService.java
```diff
@@ -58,7 +57,7 @@ public void start() {
             List$.MODULE$.<StatsFactory>empty(),
             Option.<String>empty(),
             List$.MODULE$.<Regex>empty(),
-            Map$.MODULE$.<String, CustomHttpHandler>empty(),
+            new Map.Map1<>("/prometheus", new PrometheusHandler()),
```
HenryCaiHaiying, line 60: Should this line be conditioned on whether prometheus is enabled?

src/main/java/com/pinterest/secor/monitoring/MicroMeterMetricCollector.java
```diff
@@ -19,24 +19,28 @@
...
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
 
 /**
  * MicorMeter meters can integrate with many different metrics backend 
  * (StatsD/Promethus/Graphite/JMX etc, see https://micrometer.io/docs)
  */
 public class MicroMeterMetricCollector implements MetricCollector {
+    private static final Map<String, Double> GAUGE_CACHE = new HashMap<>();
```
HenryCaiHaiying, line 42: How big will this cache become?

src/main/java/com/pinterest/secor/common/monitoring/OstrichAdminService.java
pdambrauskas: Handler will return 404, if there is no PrometheusMetricRegistry added on Micrometer (see handler class itself).
But I guess we could add a check here too, to make it more obvious while reading the code.

src/main/java/com/pinterest/secor/monitoring/MicroMeterMetricCollector.java
pdambrauskas: It depends, on how many different gauges will be tracked. Currently it will be 2*topic-count. Maybe I should limit cache size to avoid memory leak, if someone adds way too many metrics with different tags?
The cache is used, because prometheus is returning NaNs otherwise. https://stackoverflow.com/questions/50821924/micrometer-prometheus-gauge-displays-nan

HenryCaiHaiying: I suggest to add a config param for the cache size (e.g. 100), if the cache goes over that size, throw a runtime exception with error message to tell user to increase the cache size parameter.

pdambrauskas: Maybe just log warning message? We probably shouldn't crash writing process because of metric cache overflow.

HenryCaiHaiying:
In that case use the ERROR level logging and set the cache size threshold a
little bigger: 500

src/main/java/com/pinterest/secor/monitoring/MicroMeterMetricCollector.java
```diff
@@ -63,13 +78,21 @@ public void increment(String label, int delta, String topic) {
...
     public void metric(String label, double value, String topic) {
-        Metrics.gauge(label, Collections.singletonList(
-            Tag.of("topic", topic)), new AtomicDouble(0)).set(value);
+        gauge(label, value, topic);
     }
 
     @Override
     public void gauge(String label, double value, String topic) {
+        String key = label + "_" + topic;
+        if (mGaugeCache.size() <= mConfig.getMicroMeterCacheSize()) {
```
HenryCaiHaiying, line 87: I think we should only do the cache size check and issue error message when the cache needs to grow, not each time gauge is called (there will be too frequent)

src/main/java/com/pinterest/secor/monitoring/MicroMeterMetricCollector.java
```diff
@@ -63,13 +78,22 @@ public void increment(String label, int delta, String topic) {
...
+        gauge(label, value, topic);
     }
 
     @Override
     public void gauge(String label, double value, String topic) {
+        String key = label + "_" + topic;
+        if (!mGaugeCache.containsKey(key) && mGaugeCache.size() >= mConfig.getMicroMeterCacheSize()) {
+            LOG.error("Gauge cache size reached maximum, this may result in inaccurate metrics, "
+                    + "you can increase cache size by changing "
+                    + "\"secor.monitoring.metrics.collector.micrometer.cache.size\" property.");
```
HenryCaiHaiying, line 90: Can you put 'return' statement here?  Otherwise Metrics.gauge() result might be un-deterministic

HenryCaiHaiying:

PR #1482, opened 2020-07-28
Fix file extension after trim

pdambrauskas:
After trimming files, dest files were losing their extensions and using codec extension instead.
So `secor.file.extension` was ignored, after trimming files.

Also LogFilePath constructor used in tests & verifier was not working with double extentions, for exapmle `.snappy.parquet`.

### DIFF
--- a/src/main/java/com/pinterest/secor/common/LogFilePath.java
+++ b/src/main/java/com/pinterest/secor/common/LogFilePath.java
@@ -119,15 +119,15 @@ public class LogFilePath {
         // Parse basename.
         String basename = pathElements[pathElements.length - 1];
         // Remove extension.
-        int lastIndexOf = basename.lastIndexOf('.');
-        if (lastIndexOf >= 0) {
-            mExtension = basename.substring(lastIndexOf, basename.length());
-            basename = basename.substring(0, lastIndexOf);
+        int indexOfSeparator = basename.indexOf('.');
+        if (indexOfSeparator >= 0) {
+            mExtension = basename.substring(indexOfSeparator);
+            basename = basename.substring(0, indexOfSeparator);
         } else {
             mExtension = "";
         }
         String[] basenameElements = basename.split("_");
-        assert basenameElements.length == 3: Integer.toString(basenameElements.length) + " == 3";
+        assert basenameElements.length == 3: basenameElements.length + " == 3";
         mGeneration = Integer.parseInt(basenameElements[0]);
         mKafkaPartitions = new int[]{Integer.parseInt(basenameElements[1])};
         mOffsets = new long[]{Long.parseLong(basenameElements[2])};
--- a/src/main/java/com/pinterest/secor/uploader/Uploader.java
+++ b/src/main/java/com/pinterest/secor/uploader/Uploader.java
@@ -220,10 +220,8 @@ public class Uploader {
         mFileRegistry.deleteWriter(srcPath);
         try {
             CompressionCodec codec = null;
-            String extension = "";
             if (mConfig.getCompressionCodec() != null && !mConfig.getCompressionCodec().isEmpty()) {
                 codec = CompressionUtil.createCompressionCodec(mConfig.getCompressionCodec());
-                extension = codec.getDefaultExtension();
             }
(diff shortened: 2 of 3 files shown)


src/main/java/com/pinterest/secor/common/LogFilePath.java
```diff
@@ -119,7 +119,7 @@ public LogFilePath(String prefix, String path) {
         // Parse basename.
         String basename = pathElements[pathElements.length - 1];
         // Remove extension.
-        int lastIndexOf = basename.lastIndexOf('.');
```
HenryCaiHaiying, line 122: What if the file name is a.b.c while a.b is the basename and c is the extension name?

HenryCaiHaiying: The logic to calculate the file extension name is in ParitionFinalizer line 65 (also in MessageWriter, line 64), I think we should use the same logic.  Can you move that logic into a common utility routine and have all three places calling that utility routine?

src/main/java/com/pinterest/secor/uploader/Uploader.java
```diff
@@ -233,7 +231,7 @@ private void trim(LogFilePath srcPath, long startOffset) throws Exception {
                         dstPath = new LogFilePath(localPrefix, srcPath.getTopic(),
                                                   srcPath.getPartitions(), srcPath.getGeneration(),
                                                   srcPath.getKafkaPartition(), startOffset,
-                                                  extension);
+                                                  srcPath.getExtension());
```
HenryCaiHaiying, line 234: In the case that file is a.snappy.parquet, srcPath.extension =  parquet, where is '.snappy' part being captured in LogFilePath?

Or how you do have 'a.snappy.parquet' in the first place, did you specify that in SecorConfig.getFileExtension config?

pdambrauskas: We're using `secor.file.extension` config and specifying file extension. There is no other way to have double extension.
We are doing that to mimic Apache Spark file name convention, at least extensions.

MessageWriter is respecting that:
```java
if (mConfig.getCompressionCodec() != null && !mConfig.getCompressionCodec().isEmpty()) {
            mCodec = CompressionUtil.createCompressionCodec(mConfig.getCompressionCodec());
            mFileExtension = mCodec.getDefaultExtension();
        }
        if (mConfig.getFileExtension() != null && !mConfig.getFileExtension().isEmpty()) {
            mFileExtension = mConfig.getFileExtension();
        } else if (mFileExtension == null){
            mFileExtension = "";
        }
```
So I changed it to reuse same extention from path, generated by MessageWritter

src/main/java/com/pinterest/secor/common/LogFilePath.java
pdambrauskas: > What if the file name is a.b.c while a.b is the basename and c is the extension name?
Yes, that would be a problem, but in that case we this constructor will fail.
we have this on line 131.
```java
        mGeneration = Integer.parseInt(basenameElements[0]);
```
Then we'd have `Integer.parse("a.b")`, so in that case this contructor would not support both, double extensions and files with multiple dots in their name. I made this change with asumption that base name of the file is allways a parseable number without dots.


> The logic to calculate the file extension name is in ParitionFinalizer line 65 (also in MessageWriter, line 64), I think we should use the same logic. Can you move that logic into a common utility routine and have all three places calling that utility routine?

You mean I should use it on `Uploader` class instead of using `srcPath.getExtention()`? My first idea was to do that, but since we already have extension on `srcPath`, and `dstPath` is path for trimmed srcFile, I thougt it is reasonable to reuse same extension instead regenerating it from configuration.
I do agree that PartitionFinalizer and MessageWritter logics should be extracted to common utility though :)

HenryCaiHaiying: I see.  In this case, can you rename the variable name 'lastIndexOf' to 'indexOf'?

HenryCaiHaiying:
Left a small comment.

HenryCaiHaiying:
Thanks for the contribution

PR #1500, opened 2020-08-04
Adding canned ACL to uploadRequest.

clairefinnie:
Our Secor deployments are in Ops account and it is writing to a bucket in the DataScience account.

Because it is cross account, we need to set the the ACL on each object so that the bucket owner has access to the files.


clairefinnie:
Meant to be on fork repo :/

PR #1520, opened 2020-08-12
Fix building project due to insecure repo URLs

kvainer:
Maven repos now require connections over HTTPS only and won't accept HTTP


HenryCaiHaiying:
Thanks for the contribution

PR #1528, opened 2020-08-15
Looks like OstrichAdminService is looking for build.properties in com…

HenryCaiHaiying:
…/pinterest/

secor/common/monitoring folder.

So create files in that folder


HenryCaiHaiying:
Landed.

PR #1547, opened 2020-08-27
Do not fail on write error

### LINKED ISSUE #1505: UnsupportedOperationException fatal error when schema does not match
Hello, we are using Secor to parse JSON events using the JsonMessageParser, outputting to ORC files.

We encountered a fatal error the other day with an invalid event being present in the source topic, which contained a matching key name but had a different data type. The result was that the reader successfully parsed the message, but the writer failed and caused the following error:

image

It seems based on the implementation in [Consumer.java](https://github.com/pinterest/secor/blob/master/src/main/java/com/pinterest/secor/consumer/Consumer.java#L252), the intention is for Secor to skip messages that can't be parsed with a warning, but any write errors are rethrown. This caused Secor to get stuck on this one bad event and hold up the whole partition. Is this behaviour intentional or should it prevent a mismatch between reader and writer formats like this?

pdambrauskas:
Fixes #1505 

This issue occurred to us too. Found existing issue about it and made this PR.

- Moved offset parsing and message parsing to single `try/catch` block
- Reused fail backoff logic which was implemented in parse block
- Made fail threshold configurable
- Split `consumeNextMessage` to multiple smaller methods to make it more readable.

### DIFF
--- a/src/main/java/com/pinterest/secor/consumer/Consumer.java
+++ b/src/main/java/com/pinterest/secor/consumer/Consumer.java
@@ -18,7 +18,11 @@
  */
 package com.pinterest.secor.consumer;
 
-import com.pinterest.secor.common.*;
+import com.pinterest.secor.common.DeterministicUploadPolicyTracker;
+import com.pinterest.secor.common.FileRegistry;
+import com.pinterest.secor.common.OffsetTracker;
+import com.pinterest.secor.common.SecorConfig;
+import com.pinterest.secor.common.ShutdownHookRegistry;
 import com.pinterest.secor.message.Message;
 import com.pinterest.secor.message.ParsedMessage;
 import com.pinterest.secor.monitoring.MetricCollector;
@@ -34,10 +38,9 @@ import com.pinterest.secor.uploader.UploadManager;
 import com.pinterest.secor.uploader.Uploader;
 import com.pinterest.secor.util.ReflectionUtil;
 import com.pinterest.secor.writer.MessageWriter;
+import java.io.IOException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import java.io.IOException;
 
 /**
  * Consumer is a top-level component coordinating reading, writing, and uploading Kafka log
@@ -51,6 +54,7 @@ import java.io.IOException;
  */
 public class Consumer extends Thread {
     private static final Logger LOG = LoggerFactory.getLogger(Consumer.class);
+    private static final double DECAY = 0.999;
 
     protected SecorConfig mConfig;
     protected MetricCollector mMetricCollector;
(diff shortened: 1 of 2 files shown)


src/main/java/com/pinterest/secor/consumer/Consumer.java
```diff
@@ -223,37 +224,12 @@ protected boolean consumeNextMessage() {
                 }
 
                 parsedMessage = mMessageParser.parse(transformedMessage);
-                final double DECAY = 0.999;
```
HenryCaiHaiying, line 226: Did we lose the reset of DECAY after refactoring?

src/main/java/com/pinterest/secor/consumer/Consumer.java
```diff
@@ -267,4 +243,36 @@ protected boolean consumeNextMessage() {
...
+        try {
+            mMessageWriter.adjustOffset(rawMessage, isLegacyConsumer);
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to adjust offset.", e);
+        }
+    }
+
+    private void writeMessage(Message rawMessage, ParsedMessage parsedMessage) throws Exception {
+        mMessageWriter.write(parsedMessage);
+        mBadMessages *= DECAY;
```
pdambrauskas, line 257: `DECAY` is final and used to decrease `mBadMessages` as before. It does not make sense to declare new `DECAY` variable on each `writeMessage` call, so I moved it to static final constant.

pdambrauskas:
> The refactoring makes the code diff difficult to read, I wasn't sure the purpose of making those new methods. Looks like the main purpose of the PR is to add mBasMessages, you can just insert this logic in place. Also what kind of errors do you see during messageWriter.write and you want to skip those errors?

Hey Henry, thanks for taking time to review my  Pull Requests :). I do understand that it is hard to review this code, since the method was not that small.

`mBadMessages` was introduced to make it possible to ignore all failing messages, but it is not the main purpose of this PR. 

To simplify the idea, before code looked like this:
```java
//...
try {
  // parse partition
} catch (Exception e) {
  // increment consumer.message_errors.count metric
  // handle fail, re-throw if there were > 1000 errors 
}

try {
  // write message
} catch (Exception e) {
  // just re-throw
}
```
After this change there is only one `try` block:

```java
try {
  // parse partition
  // write message
} catch (Exception e) {
  // increment consumer.message_errors.count metric
  // handle fail, re-throw if there were > 1000 errors 
}
```
This allows us to configure secor to not fail on writes and also `onsumer.message_errors.count` is more correct.

> Also what kind of errors do you see during messageWriter.write and you want to skip those errors?

Our error is very similar to one, described in #1505. We use custom `ReaderWriterFactory` implementation with custom Writer to write json messages to parquet and validate schema, so exception is a bit different though, I thought it would make sense to catch exception in our Writer, but then I saw that others have the same problem on #1505. And `consumer.message_errors.count` metric is increased on Consumer class, so i decided to keep on place responsible for handling this kind of errors.

HenryCaiHaiying:
Your explanation makes sense, I will merge in your PR, thanks.

PR #1738, opened 2020-12-03
Compression Codec for Avro Writer

richiesgr:
The current config only includes compression for the file produced however Avro have hist own compression codec who is not configurable using the current config

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,4 @@ bin/
 .project
 .classpath
 .m2
+dv-deploys
--- a/src/main/config/secor.common.properties
+++ b/src/main/config/secor.common.properties
@@ -519,6 +519,10 @@ parquet.enable.dictionary=true
 # Parquet 1.9.
 parquet.validation=false
 
+# Avro doesn't use the hadoop codec for compression it includes his own (see here: org.apache.avro.file.CodecFactory)
+# value can be null(no compression),deflate,snappy,bzip2
+avro.compression.codec=null
+
 # User can configure ORC schema for each Kafka topic. Common schema is also possible. This property is mandatory
 # if DefaultORCSchemaProvider is used. ORC schema for all the topics should be defined like this:
 secor.orc.message.schema.*=struct<a:int\,b:int\,c:struct<d:int\,e:string>\,f:array<string>\,g:int>
--- a/src/main/java/com/pinterest/secor/io/impl/AvroFileReaderWriterFactory.java
+++ b/src/main/java/com/pinterest/secor/io/impl/AvroFileReaderWriterFactory.java
@@ -26,6 +26,7 @@ import com.pinterest.secor.io.FileReaderWriterFactory;
 import com.pinterest.secor.io.FileWriter;
 import com.pinterest.secor.io.KeyValue;
 import com.pinterest.secor.util.AvroSchemaRegistryFactory;
+import com.pinterest.secor.util.CompressionUtil;
 import com.pinterest.secor.util.ParquetUtil;
 import org.apache.avro.AvroRuntimeException;
 import org.apache.avro.Schema;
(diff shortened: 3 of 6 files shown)


.gitignore
```diff
@@ -44,3 +44,4 @@ bin/
 .project
 .classpath
 .m2
+dv-deploys
```
HenryCaiHaiying, line 47: What is this line for?

src/main/java/com/pinterest/secor/io/impl/AvroFileReaderWriterFactory.java
```diff
@@ -117,23 +122,21 @@ public void close() throws IOException {
...
             writer.create(schema, file);
         }
 
-        private CodecFactory getCodecFactory(CompressionCodec codec) {
-            CompressionCodecName codecName = CompressionCodecName
-                    .fromCompressionCodec(codec != null ? codec.getClass() : null);
+        private CodecFactory getCodecFactory(String avroCodec) {
             try {
-                return CodecFactory.fromString(codecName.name().toLowerCase());
+                return CodecFactory.fromString(avroCodec);
```
HenryCaiHaiying, line 139: Looks like you can have two edge cases: avroCode = "null" and avroCodec = null, maybe you need to deal with them separately

src/main/config/secor.common.properties
```diff
@@ -519,6 +519,10 @@ parquet.enable.dictionary=true
 # Parquet 1.9.
 parquet.validation=false
 
+# Avro doesn't use the hadoop codec for compression it includes his own (see here: org.apache.avro.file.CodecFactory)
+# value can be null(no compression),deflate,snappy,bzip2
+avro.compression.codec=null
```
HenryCaiHaiying, line 524: 'null' as default is unconventional, 'none' might be better.

HenryCaiHaiying (commented):
Looks good, left a few comments.

.gitignore
```diff
@@ -43,4 +43,4 @@ bin/
 .settings/
 .project
 .classpath
-.m2
+.m2
```
HenryCaiHaiying, line 46: Why this line change?

src/main/java/com/pinterest/secor/io/impl/AvroFileReaderWriterFactory.java
HenryCaiHaiying: Do you need to deal with avroCode == null || acvoCode.equalsIgnoreCase("none") here before calling CopdecFactory.fromSring?

HenryCaiHaiying (commented):
Left two comments.

.gitignore
richiesgr: Add by error removed now

src/main/java/com/pinterest/secor/io/impl/AvroFileReaderWriterFactory.java
richiesgr: It's check in the CompressionUtil class

src/main/config/secor.common.properties
richiesgr: switched to "none" string

src/main/java/com/pinterest/secor/util/CompressionUtil.java
```diff
@@ -37,9 +38,21 @@
...
+    /**
+     * Retrieve Avro compression codec from config in param : avro.compression.codec
+     * List of default supported avro compression code
+     * @see org.apache.avro.file.CodecFactory
+     *
+     * @param config secor configuration
+     * @return codec as a String or null (No compression if not set)
+     */
+    public static String getAvroCompressionCodec(SecorConfig config) {
+        return (config.getAvroCompressionCodec() != null && config.getAvroCompressionCodec().toLowerCase() == "none") ? null : config.getAvroCompressionCodec();
```
HenryCaiHaiying, line 56: This coding is a bit weird.  It will be easier to read if coded like this:

return config.getAvro() == null || config.getAvro.equalsIgnoreCase("none") ? null : config.getAvro();

Also how does the caller deal with with the null return?

src/main/java/com/pinterest/secor/util/CompressionUtil.java
```diff
@@ -37,9 +38,21 @@
...
+    }
+}
```
HenryCaiHaiying, line 58: You have some trailing whitespaces

.gitignore
HenryCaiHaiying, line 46: You might have a trailing whitespace on this line

PR #1801, opened 2021-01-07
Remove writing to Zookeeper when kafka is selected as offset storage and dual.commit is false.

Wosin:
This PR changes the way secor uses Zookeeper for storing offsets. Currently, secor will use zookeeper to store offsets no matter what settings are selected. This PR makes secor only use zookeeper for storing offsets in two cases:
 1. LegacyKafkaConsumer is used ( it will always use zookeeper for that).
 2. New KafkaConsumer is used and offset storage is set to `zookeeper` or `dual.commit` is enabled.

### DIFF
--- a/src/main/java/com/pinterest/secor/reader/LegacyKafkaMessageIterator.java
+++ b/src/main/java/com/pinterest/secor/reader/LegacyKafkaMessageIterator.java
@@ -18,11 +18,14 @@
  */
 package com.pinterest.secor.reader;
 
+import com.google.common.collect.ImmutableMap;
 import com.pinterest.secor.common.SecorConfig;
 import com.pinterest.secor.common.TopicPartition;
 import com.pinterest.secor.message.Message;
 import com.pinterest.secor.timestamp.KafkaMessageTimestampFactory;
 import com.pinterest.secor.util.IdUtil;
+import kafka.common.OffsetAndMetadata;
+import kafka.common.TopicAndPartition;
 import kafka.consumer.Blacklist;
 import kafka.consumer.Consumer;
 import kafka.consumer.ConsumerConfig;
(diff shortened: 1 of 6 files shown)


Wosin:
Tests currently fail because I've disabled possibility of using `LegacyKafkaMessageIterator` without storing offsets in zookeeper i.e. with settings `offset.storage=kafka` and `dual.commit.enabled=false`. It is because the old KafkaConsumer does not allow us to read the committed offsets for the current partition and we need zookeeper to do that.

src/main/java/com/pinterest/secor/uploader/Uploader.java
```diff
@@ -171,15 +176,21 @@ protected void uploadFiles(TopicPartition topicPartition) throws Exception {
                 if (mDeterministicUploadPolicyTracker != null) {
                     mDeterministicUploadPolicyTracker.reset(topicPartition);
                 }
-                mZookeeperConnector.setCommittedOffsetCount(topicPartition, lastSeenOffset + 1);
```
HenryCaiHaiying, line 174: Why this line is removed?

src/main/java/com/pinterest/secor/reader/LegacyKafkaMessageIterator.java
```diff
@@ -99,7 +102,11 @@ public void init(SecorConfig config) throws UnknownHostException {
 
     @Override
     public void commit(TopicPartition topicPartition, long offset) {
-        mConsumerConnector.commitOffsets();
+        TopicAndPartition kafkaTopicPartition = new TopicAndPartition(topicPartition.getTopic(), topicPartition.getPartition());
+        kafka.common.OffsetAndMetadata offsetAndMetadata = OffsetAndMetadata.apply(offset);
+
+        LOG.info("committing {} offset {} to kafka", topicPartition, offset);
+        mConsumerConnector.commitOffsets(ImmutableMap.of(kafkaTopicPartition, offsetAndMetadata), true);
```
HenryCaiHaiying, line 109: What does the argument 'true' mean?

HenryCaiHaiying (commented):
Thanks for the contribution.  Mostly looks fine, left two comments.

For the failed tests, can you disable the tests on legacy path with zookeeper storage = false?

HenryCaiHaiying:
@Wosin There are a few comments in the PR

src/main/java/com/pinterest/secor/uploader/Uploader.java
Wosin: It wasn't. It was wrapped into `if` statement and now currently in lines 185-189.

src/main/java/com/pinterest/secor/reader/LegacyKafkaMessageIterator.java
Wosin: It means that if the offsets commit will fail, there will be no retries after failing to commit offsets, which is consistent with the previous behaviour, since `mConsumerConnector.commitOffsets()` did the same thing.

HenryCaiHaiying:
for the test failures, is that possible to skip those tests if the combination no longer makes sense?

Wosin:
Disabled that + added the failfast behaviour, since in the previous case it would only fail on upload which doesn't make sense :)

HenryCaiHaiying:
Looks good, I will merge it in.  Thanks for the contribution.

Issue #1949, opened 2021-03-26
Expose JVM metrics when using Micrometer


DomWos:
Currently, only the custom metrics are exposed when using micrometer. Adding JVM internal metrics can be done in few lines of code and can greatly help when running secor on production.

HenryCaiHaiying:
That sounds a good idea.  Will you be able to make a PR to make the
change?  It would be a great help for the community.

DomWos:
Sure, will do it over the weekend :)

HenryCaiHaiying:
Thanks

PR #1953, opened 2021-03-27
Expose system metrics when using micrometer.

Wosin:
Closes https://github.com/pinterest/secor/issues/1949


HenryCaiHaiying:

PR #2065, opened 2021-05-14
Add handler logic when a bad message occurs

jvaesteves:
Here at OLX Brasil, we use Secor to ingest data from Kafka to our Data Lake, but in case of an exception during the parsing of the message, we cannot ignore it. So this PR add the possibility to send every bad message to a new topic on Kafka, where the source can investigate what could have happened.


.gitignore
```diff
@@ -44,3 +44,4 @@ bin/
 .project
 .classpath
 .m2
+.vscode/
```
HenryCaiHaiying, line 47: This should be removed

src/main/java/com/pinterest/secor/common/SecorConfig.java
```diff
@@ -273,6 +273,14 @@ public int getMaxBadMessages() {
         return getInt("secor.consumer.max_bad_messages", 1000);
     }
 
+    public boolean getBadMessagesTopicEnabled() {
+        return getBoolean("secor.consumer.bad_message.topic.enabled", false);
+    }
+
+    public String getBadMessagesTopicSuffix() {
+        return getString("secor.consumer.bad_message.topic.suffix", "_BAD_MESSAGES");
```
HenryCaiHaiying, line 281: Instead of defaulting, we should throw an error if this property is not specified.

src/main/java/com/pinterest/secor/error/SendToKafkaBadMessageHandler.java
```diff
@@ -0,0 +1,70 @@
...
+            null,
+            null,
+            message.getKafkaKey(),
+            message.getPayload(),
+            messageHeaders
+        );
+
+        return mKafkaProducer.send(record, (recordMetadata, kafkaException) -> {
+            if (kafkaException != null) {
+                LOG.trace("Failed to send message to Kafka: {}", message, kafkaException);
```
HenryCaiHaiying, line 54: This should be a warning logging

HenryCaiHaiying (commented):
Thanks for this contribution.  On the high level I have one concern if there are lots of errors/exceptions, there will be many kafka message to be produced to bad_messages topic, this might affect the performance of consumer.  An alternative design is to have a separate thread to send the message and use a queue to communicate between consumer and producer.   Or add a sampling feature which you only send some percentage of messages.

Issue #2155, opened 2021-11-16
PartitionFinalizer fails when there is no data produced in the topic eventhough there are other topics with data


yesemsanthoshkumar:
Secor: 0.29
Kafka: 2.4.1.1 (AWS MSK)

Hi,
We are experimenting with secor for storing kafka messages in s3. We have topics subscribed to secor through `secor.kafka.topic_filter` config. The filtered topics are of 2 types. One that gets data always in that topics. The other wouldn't get data for a significant duration. The problem we are encountering is with the latter. When there is no data in kafka for that particular topic, the partition finalizer when run for the hourly partitions fails saying the offset seek cannot be a negative number.

## Stacktrace
```
2021-11-16 12:35:33,619 [main] (com.pinterest.secor.main.PartitionFinalizerMain:49) ERROR Partition finalizer failed
java.lang.IllegalArgumentException: seek offset must not be a negative number
	at org.apache.kafka.clients.consumer.KafkaConsumer.seek(KafkaConsumer.java:1541)
	at com.pinterest.secor.common.SecorKafkaClient.getLastMessage(SecorKafkaClient.java:69)
	at com.pinterest.secor.parser.PartitionFinalizer.getFinalizedUptoPartitions(PartitionFinalizer.java:83)
	at com.pinterest.secor.parser.PartitionFinalizer.finalizePartitions(PartitionFinalizer.java:212)
	at com.pinterest.secor.main.PartitionFinalizerMain.main(PartitionFinalizerMain.java:47)
```
The particular topic that throwed this error has fewer records and all have been processed. Later there was no data for that particular topic and its partitions when the finalizer ran.

## What I found (in the order of occurence):
1. When there is no data in the partition, `endOffsets` method on kafkaConsumerAPI gives out 0 as offset.
(https://kafka.apache.org/22/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#endOffsets-java.util.Collection-) 

2. And in secor we are doing seek on the obtained endOffset value minus 1 which results to -1 to the `seek` method of kafkaConsumer (https://github.com/pinterest/secor/blob/v0.29/src/main/java/com/pinterest/secor/common/SecorKafkaClient.java#L69).

3. And this -1 in seek is throwing the above exception (https://kafka.apache.org/22/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#seek-org.apache.kafka.common.TopicPartition-long-).

## Question
[...]

PR #2156, opened 2021-11-20
Add guard for consumer.seek when using endoffsets

yesemsanthoshkumar:
address #2155 
When there is no message written in the topic partition, the endOffsets method return 0 as default.
https://kafka.apache.org/22/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#endOffsets-java.util.Collection-
This causes seek value to be -1 which fails the PartitionFinalizer since the offset is negative.
https://kafka.apache.org/22/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#seek-org.apache.kafka.common.TopicPartition-long-


yesemsanthoshkumar:
@HenryCaiHaiying Please review.

HenryCaiHaiying:
Looks fine.  Can you use this coding style?

if (cond) {
    stmt;
}

yesemsanthoshkumar:
Updated.

HenryCaiHaiying:
Thanks for the contribution

